PR #13, opened 2022-09-14
refactor: modular layers

alexander-camuto:
- Simplified layer interfaces whereby both Affine and Conv layers now ingest "Kernels" which can have fixed or advice parameters. 
- Both Affine and Conv layers use the same "Input/Output" layer types (rather than a split between input and image). 
- Inputs, Outputs and Parameters use the same IOConfig to assign and layout


alexander-camuto:
TODO: add bias back in to affine layers

PR #22, opened 2022-09-22
refactor: fully dynamic conv layers

### LINKED ISSUE #18: Fully dynamic layers
- Affine layers now pull their dimensionality from Val/VarTensors. We should apply this as much as possible to Envwise operations and Convolutional layers (some parameters like a convolutional layer's Stride and Padding can't be extracted from inputs).

alexander-camuto:
- removes input channels as generic on conv layers using:
    -   4D kernels convolved with 3D inputs in `tensor_ops::convolution`
    - allowing for N-d vectors on IOconfig (i.e N-d advice vectors, N-d inputs, N-d kernels)
    - removing `[1,len]` shape requirement for 1-D input

**NOTE:** Swaps width and Height dims to be closer to `pytorch` norm

Resolves #18

### DIFF
--- a/examples/conv2d_mnist/main.rs
+++ b/examples/conv2d_mnist/main.rs
@@ -163,18 +163,18 @@ where
         );
 
         let l0q: EltwiseConfig<F, BITS, DivideBy<F, 32>> =
-            EltwiseConfig::configure(cs, advices.get_slice(&[0..LEN], &[1, LEN]), None);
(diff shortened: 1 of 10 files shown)


jasonmorton:
It seems like the conv proof is a bit slower now (679s vs 624s on my machine).   Perhaps because of the layout change; it doesn't seem like it should be the dynamic part.  Is that what you see as well?
```
VK took 55
PK took 43
Proof took 679
Verify took 4
```
vs 
```
VK took 56
PK took 37
Proof took 638
Verify took 4
```

PR #27, opened 2022-09-26
feat: add 2D max pooling to tensor ops

horacepan:
Implement max pool 2d + rustdoc

Implementation of max pool 2D is pretty much follows the same structure as `convolution`. I'm not entirely sure about using `into_iter` but I think it should be okay since `get_slice` clones the underlying data anyhow. Happy to make any changes/modify if you guys think it should be done a different way.

### DIFF
--- a/src/tensor_ops/mod.rs
+++ b/src/tensor_ops/mod.rs
@@ -1,6 +1,7 @@
 pub mod eltwise;
 use crate::tensor::{Tensor, TensorType};
 pub use std::ops::{Add, Mul};
+pub use std::cmp::Ord;
 
 pub fn matmul<T: TensorType + Mul<Output = T> + Add<Output = T>>(
     kernel: Tensor<T>,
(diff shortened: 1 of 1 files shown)


alexander-camuto:
Hey Horace the function looks really good and super clean ! 

I quickly tested the code on an input composed of Halo2's `Value` types (which we might expect to encounter when building the circuit), and the example fails because `Value` does not implement the `Ord` trait the function requires, so we might need to get creative on how we implement the `max()` operation on those types. 

To reproduce the issue you can switch out the  (much appreciated) docstring example to: 

```rust
use halo2deeplearning::tensor::Tensor;
use halo2deeplearning::tensor_ops::max_pool2d;
use halo2curves::pasta::Fp as F;
use  halo2_proofs::circuit::Value;

let x: Tensor<Value<F>> = Tensor::<i32>::new(
    Some(&[5, 2, 3, 0, 4, -1, 3, 1, 6]),
     &[1, 3, 3]).unwrap().into();
const PADDING: usize = 0;
const POOL_H: usize = 2;
const POOL_W: usize = 2;
const STRIDE: usize = 1;
let pooled = max_pool2d::<Value<F>, PADDING, POOL_H, POOL_W, STRIDE>(x);
let expected: Tensor<Value<F>> = Tensor::<i32>::new(Some(&[5, 4, 4, 6]), &[1, 2, 2]).unwrap().into();
assert_eq!(pooled, expected);

```

horacepan:
ah good catch @alexander-camuto, I completely missed that. Hmm, yea will have to think a bit about that a bit ...

horacepan:
abandoning this pr

PR #28, opened 2022-09-27
chore: documentation

### LINKED ISSUE #19: Documentation
Currently only `Tensor` has doc strings. As we start to cement the structure of types we should look to apply the same treatment to other structs and traits.

alexander-camuto:
- Removes `IOConfig` entirely (realized when writing docs that it was superfluous). 
- resolves #19.


jasonmorton:
As discussed this folding of IOConfig into VarTensor makes the lib easier to understand, and should also make it easier to implement layout strategy changes (e.g. one column vs. many columns) that are invisible to the user of VarTensor.

PR #32, opened 2022-10-05
feat: add 2D max pooling to tensor ops

horacepan:
- Adds a `tmax` function to the `TensorType` trait, which returns the larger of the two inputs
- the `max_pool2d` op computes the maximum for each slice by calling `fold` over the slice with `tmax`

Concerns:
- I'm not entirely sure about what operations are/aren't allowed for `AssignedCell`, `Assigned`, `Value` types.
- For the `f32` behavior, I had the `tmax` return the non-NAN value one of the operands of `tmax` was `f32::NAN`.
- for `pasta::Fp` values, should we handle negative numbers explicitly in the `tmax` function?

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -22,11 +22,16 @@ use std::iter::Iterator;
 use std::ops::Deref;
 use std::ops::DerefMut;
 use std::ops::Range;
+use std::cmp::max;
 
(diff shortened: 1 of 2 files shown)


alexander-camuto:
@horacepan looks great -- merging into another branch where we've implemented a slew of other basic ops. Thanks for all your help !

PR #33, opened 2022-10-05
fix: conv rescaling and unknown values for fixed tensors

alexander-camuto:
- fixes conv `.onnx` examples which would fail due to scaling issues
- creates a generic framework for rescaling layers
- creates a generic framework for rescaling element-wise activations
- fixes unknown value tensors during full proof generation for fixed variables (all examples)


alexander-camuto:
Todo: still need to fix the max advice calculation for ONNX models

PR #35, opened 2022-10-07
fix: sigmoid OOR issues

alexander-camuto:
- Introduces a simple method for tracking quantization level / denominator throughout a network 
- Correctly quantizes sigmoid inputs and outputs
- Fixed selector multiply by 0 issue for sigmoid by introducing a `default_value` method on the Non-linearity trait
- CLI can now adjust lookup table bits and the quantization denom


jasonmorton:
Let's set a default for bits and scale, and make sure all the example onnx files are run in the tests.

PR #50, opened 2022-10-29
chore: helpful error messages for mock proving

alexander-camuto:
- Gives user suggestions on how to fix common OOR Lookup or Permutation errors. 
- `debug` logs give additional information to those errors

eg. 
image


jasonmorton:
Looks good but we may still want to pass on the halo2 errors for debugging.

PR #63, opened 2022-11-18
refactor: struct for tracking variables

alexander-camuto:
- Creates a new struct for holding advice, fixed, and instance columns. 
- First step in creating a modular interface for flipping public / private variables.


jasonmorton (commented):
We should rename Chip since it is already a halo2 trait.  Perhaps ColumnManager?

PR #72, opened 2022-12-05
add: leaky_relu non-linearity

dcbuild3r:
Add Leaky ReLU by rebasing from main


src/graph/node.rs
```diff
@@ -309,6 +313,21 @@ impl Node {
...
+                mn.out_dims = input_node.out_dims.clone();
+                mn.output_max = input_node.output_max;
+                mn.in_scale = input_node.out_scale;
+                mn.out_scale = scale;
+                let scale_diff = mn.in_scale - mn.out_scale;
+                // We can also consider adjusting the scale of all inputs and the output in a more custom way.
+                if scale_diff > 0 {
+                    let mult = scale_to_multiplier(scale_diff);
+                    mn.opkind = OpKind::LeakyReLU(mult as usize); // now the input will be scaled down to match
+                    mn.output_max = input_node.output_max / mult;
```
alexander-camuto, line 328: We also want to read in the leaky relu slope somewhere here; from the Onnx node.

alexander-camuto:
Hey @dcbuild3r thanks for this its looking really good. 

In terms of next steps: 
- [ ] ensure non-linearity behaves as expected by generating an Onnx graph / input file for a graph with a single LeakyReLU op (with slope 0.05) and then running ezkl on it
- [ ] once it is working you can add that example to [this repo](https://github.com/zkonduit/onnx-examples) and add the example to the `tests/integration_tests.rs` file. 
- [ ] more broadly we probably want to think about how to generalize ingesting the slope parameter into the LeakyReLU function. It could just be item 2 of the `scale` array or something. Concurrently we also need a way to load  that parameter from the Onnx node. I've added a comment where it should probably happen in the code. Onnx / tract can be kind of fiddly and have inconsistent interfaces for loading parameters so lmk if you need help with that.

PR #82, opened 2022-12-19
feat: make max_rotations configurable

### LINKED ISSUE #78: Flag for setting max number of used rows
Max number of "used" rows is currently hardcoded to `512`. We should make this value settable using a cli flag /  config field (possibly as part of `.json` config file update).

Genysys:
This PR makes the tensor max value configurable via a CLI arg

closes #78

### DIFF
--- a/benches/affine.rs
+++ b/benches/affine.rs
@@ -33,10 +33,10 @@ impl<F: FieldExt + TensorType> Circuit<F> for MyCircuit<F> {
     fn configure(cs: &mut ConstraintSystem<F>) -> Self::Config {
         let len = unsafe { LEN };
 
-        let input = VarTensor::new_advice(cs, K, len, vec![len], true);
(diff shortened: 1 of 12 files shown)


alexander-camuto:
Hey @Genysys  and @samtvlabs this looks great ! Added a few nitpicky comments re-naming and comments, otherwise looks great. Once those changes are made will run integration tests ☺️

jasonmorton:
This is great and happy to have your contributions.  I think Dante's comments are not visible.  The main thing is that we should change the naming of this feature perhaps to "max_rotations," as that is how it is currently used (it is not the max tensor value or the max rows, since there may be more than one block of this size stacked vertically).

alexander-camuto (changes requested):
Hey @Genysys and @samtvlabs this looks great ! Added a few nitpicky comments re-naming and comments, otherwise looks great. Once those changes are made will run integration tests ☺️

alexander-camuto (approved):
LGTM !

PR #110, opened 2023-02-18
refactor: exact instances of Option / Result combinators

huitseeker:
Tool-aided by [comby-rust](https://github.com/huitseeker/comby-rust)

Opinions on this one loosely held, I’m happy to adapt.


alexander-camuto (approved):
LGTM ! pending tests

PR #113, opened 2023-02-24
Minor adjustment to the Solidity verify() function signature and integration of Solidity verification with the CI

weijiekoh:
This PR does two main things:

1. Change the ABI of the `verify()` function in Solidity to:

```solidity
    function verify(
        uint256[] memory pubInputs,
        bytes memory proof
    ) public view returns (bool)
```
Previously, the size of `pubInputs` had to be specified. This change makes it easier for the CI tests to run as the number of inputs is not known at compile-time.

2. Enable Solidity verification in the CI

The CI flow now includes verification of the generated Solidity verifiers for the models which can be verified in Solidity (the others create contracts that are too large to work).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ tensorflow = {version = "0.18.0", features = ["eager"], optional = true }
 plotters = { version = "0.3.0", optional = true }
 tract-onnx = { version = "0.17.7", optional = true }
 anyhow = "1.0.65"
-clap = { version = "4.0.7", features = ["derive"] }
+clap = { version = "4.0.32", features = ["derive"] }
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -14,7 +14,7 @@ tensorflow = {version = "0.18.0", features = ["eager"], optional = true }
 plotters = { version = "0.3.0", optional = true }
 tract-onnx = { version = "0.17.7", optional = true }
 anyhow = "1.0.65"
-clap = { version = "4.0.7", features = ["derive"] }
+clap = { version = "4.0.32", features = ["derive"] }
```
weijiekoh, line 17: This change had to be made to resolve an error where some other package needed a different version of a shared dependency.

PR #114, opened 2023-02-24
chore: minor adjustment to the Solidity verify() function signature and integration of Solidity verification with the CI

weijiekoh:
This PR does two main things:

1. Change the ABI of the `verify()` function in Solidity to:

```solidity
    function verify(
        uint256[] memory pubInputs,
        bytes memory proof
    ) public view returns (bool)
```
Previously, the size of `pubInputs` had to be specified. This change makes it easier for the CI tests to run as the number of inputs is not known at compile-time.

2. Enable Solidity verification in the CI

The CI flow now includes verification of the generated Solidity verifiers for the models which can be verified in Solidity (the others create contracts that are too large to work).

### DIFF
--- a/src/execute.rs
+++ b/src/execute.rs
@@ -23,7 +23,6 @@ use halo2_proofs::poly::VerificationStrategy;
 use halo2_proofs::transcript::{Blake2bRead, Blake2bWrite, Challenge255};
 use halo2_proofs::{dev::MockProver, poly::commitment::ParamsProver};
 use halo2curves::bn256::{Bn256, Fr, G1Affine};
-use halo2curves::group::ff::PrimeField;
 use log::{info, trace};
 use snark_verifier::loader::native::NativeLoader;
 use snark_verifier::system::halo2::transcript::evm::EvmTranscript;
@@ -34,11 +33,7 @@ use thiserror::Error;
 use std::fs::File;
 use std::io::Write;
 use std::process::Command;
-use ethers_solc::Solc;
-use ethers::contract::ContractFactory;
-use ethers::contract::abigen;
-use ethers::types::U256;
-use super::eth::setup_eth_backend;
+use super::eth::verify_proof_via_solidity;
 
 /// A wrapper for tensor related errors.
 #[derive(Debug, Error)]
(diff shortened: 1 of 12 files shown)


src/execute.rs
```diff
@@ -422,7 +432,38 @@ pub fn run(cli: Cli) -> Result<(), Box<dyn Error>> {
...
+                    for val in &proof.instances[0] {
+                        let bytes = val.to_repr();
+                        let u = U256::from_little_endian(bytes.as_slice());
+                        public_inputs.push(u);
+                    }
+
+                    let result = contract.verify(
+                        public_inputs,
+                        ethers::types::Bytes::from(proof.proof.to_vec()),
+                    ).call().await.unwrap();
```
alexander-camuto, line 460: Can we box some of these errors and return them via (?)

src/execute.rs
```diff
@@ -422,7 +432,38 @@ pub fn run(cli: Cli) -> Result<(), Box<dyn Error>> {
...
+
+                    info!("Solidity verification result: {}", result);
+                    assert!(result);
```
alexander-camuto, line 463: maybe box this up in a similar fashion so we don't get hard failures if used by `pyezkl`. This could for instance be similar to the error handling in the `evm_verify` function: 

```rust
 if result.reverted {
            return Err(Box::new(EvmVerificationError::Reverted));
        }
```

Though I'm not sure this result object will be the same -- so the above might have to be tailored to the returned object

src/execute.rs
```diff
@@ -422,7 +432,38 @@ pub fn run(cli: Cli) -> Result<(), Box<dyn Error>> {
             ProofSystem::KZG => {
                 let proof = Snark::load::<KZGCommitmentScheme<Bn256>>(&proof_path, None, None)?;
                 let code = DeploymentCode::load(&deployment_code_path)?;
-                evm_verify(code, proof)?;
+                evm_verify(code, proof.clone())?;
+
+                if sol_code_path.is_some() {
+                    let (anvil, client) = setup_eth_backend().await;
```
alexander-camuto, line 438: if you  want to -- you can bundle up this section as a standalone function -- whatever your preference is. It might be useful to do so later on; especially if `pyezkl` calls components of this

alexander-camuto (changes requested):
Added a few small comments regarding error handling, otherwise looks perfect !

src/execute.rs
weijiekoh: :+1: done, check out https://github.com/zkonduit/ezkl/commit/d6abd3f6b5614364437a8cd83f25a1c903f8edfd#diff-b9cde71f35f4266fc72cee1dccc59f632255fa628e923d8cbd9c6429b7a1dcf6R46

src/execute.rs
weijiekoh: Sure! Check out https://github.com/zkonduit/ezkl/commit/d6abd3f6b5614364437a8cd83f25a1c903f8edfd#diff-b9cde71f35f4266fc72cee1dccc59f632255fa628e923d8cbd9c6429b7a1dcf6R75

src/execute.rs
weijiekoh: Ditto, check out https://github.com/zkonduit/ezkl/commit/d6abd3f6b5614364437a8cd83f25a1c903f8edfd#diff-b9cde71f35f4266fc72cee1dccc59f632255fa628e923d8cbd9c6429b7a1dcf6R75

`return Err(Box::new(EvmVerificationError::InvalidProof));`

alexander-camuto (approved):
LGTM !

PR #118, opened 2023-02-25
fix: Update README

lancenonce:
Updated README to:

- Change target/realease/ezkl tags
- Split up Bash scripts
- Added script for increasing SRS Row size


alexander-camuto (approved):
LGTM ! thank you

PR #137, opened 2023-03-01
fix: tighter packing of columns

alexander-camuto:
Currently we use the maximum node size to upper bound the size of each variable in the circuit. This creates some very inefficient usage of advice / fixed columns within circuit. For instance: 

image0

Its high time we created more efficient circuit packing ! Here we determine the maximum size of each variable at a given _index_ rather than across the entire circuit. This creates a much tighter circuit layout. For the same circuit: 
4l

This should reduce our memory requirements significantly !


jasonmorton:
Just want to acknowledge @danilowhk for helping to diagnose this issue!

PR #156, opened 2023-03-08
feat: shared lookup config 

alexander-camuto:


jasonmorton:
As hoped this is a significant performance boost (e.g. twice as fast on the test suite), with a more significant gain for bigger models. We probably want to make --single-lookup the default.

PR #171, opened 2023-03-21
chore: add associated gen.py files in the onnx model directories (examples) wherever missing + tanh activation

disirulla:
Couldn't understand the export function in the existing gen.py files. Hence added the files without the export function


jasonmorton:
It would be better to split this into two PRs, one for tanh (feat: add tanh nonlinearity) and one for the gen.py (chore: ...) . This will make it easier to debug the tests that are failing (and is good practice).

PR #197, opened 2023-04-20
feat: Add python bindings for mock and prove

JSeam2:
Changes:
1. Add python bindings for mock and prove
2. Refactor the run args in the function signature into a py_run_args class for better usability.
3. Adds python type conversions for VarVisibility, CheckMode, TranscriptType, StrategyType
4. Created a to_var_visibility helper function


src/python.rs
```diff
@@ -164,9 +186,176 @@ fn forward(
...
+    let logrows = run_args.logrows;
+    let check_mode = run_args.check_mode;
+    let data = prepare_data(data);
+    let visibility = run_args.to_var_visibility();
+
+    match data {
+        Ok(d) => {
+            let procmodel = Model::<Fr>::new(model, run_args, Mode::Prove, visibility);
+
+            match procmodel {
```
alexander-camuto, line 277: Could we clean this up and replace each match with something like 
```rust 
 let procmodel = Model::<Fr>::new(model, run_args, Mode::Prove, visibility);
 // replace with 
  let procmodel = Model::<Fr>::new(model, run_args, Mode::Prove, visibility).map_err(|_| PyIOError::new_err("Failed to process model"))?; 
```

src/python.rs
```diff
@@ -164,9 +186,176 @@ fn forward(
...
+    let run_args: RunArgs = py_run_args.unwrap_or_else(PyRunArgs::new).into();
+    let logrows = run_args.logrows;
+    let data = prepare_data(data);
+    let visibility = run_args.to_var_visibility();
+
+    match data {
+        Ok(d) => {
+            let procmodel = Model::<Fr>::new(model, run_args, Mode::Mock, visibility);
+
+            match procmodel {
```
alexander-camuto, line 209: Could we clean this up and replace each match with something like 
```rust 
 let procmodel = Model::<Fr>::new(model, run_args, Mode::Prove, visibility);
 // replace with 
  let procmodel = Model::<Fr>::new(model, run_args, Mode::Prove, visibility).map_err(|_| PyIOError::new_err("Failed to process model"))?; 
```

alexander-camuto (changes requested):
LFG ! Added a few comments to reduce the pyramid of matches and keep things clean :)

src/python.rs
JSeam2: Let me try this, will get back soon!

alexander-camuto (approved):
SEND IT

PR #198, opened 2023-04-20
refactor: TypeModel for Onnx

### LINKED ISSUE #189: InstanceNorm support
"InstanceNorm" is not currently supported and prompts an node error.

It's similar to batch norm but slightly different—mean and variance are calculated at inference time on the input.

batch norm: https://github.com/onnx/onnx/blob/main/docs/Operators.md#BatchNormalization

instance norm: https://github.com/onnx/onnx/blob/main/docs/Operators.md#InstanceNormalization

alexander-camuto:
We currently use an `InferenceModel` as a type when loading the Onnx Model. By specifying just a few inputs and calling `into_typed` and `into_decluttered` on the `InferenceModel` we can get a `TypedModel`. Which: 

- No longer has const nodes. 
- Decomposes more complex ops into a subset of simpler operations eg. Affine = MM + ADD. This means we get new operations for free, including `InstanceNorm`. This PR thus also resolves #189. 
- Has fully specified shape information for all nodes. 

This greatly update introduces much more consistency in the way we load Onnx ops -- as seen by the now reused `load_axis_op` and `load_unary_op`  functions. 

The removal of the const nodes also reduces code footprint: we no longer need multiple quantization passes to make sure nodes are at the right scale. 

The `Op` trait as a result is much less heavy and only has a few methods. 

Massive shoutout to @kali for helping out with this.


kali:
be careful with "No longer has const nodes." This will not be true moving ahead. As a matter of fact, between tract 0.19 and 0.20 there are huge changes in the operator set. Most of the Unary nodes (which are Binary with Const embedded) are removed and replaced by just that Binary + Const. In the long run, it reduces the Operator set even more.

0.19 -> 0.20 is the biggest change to tract-core operator set I have ever done, so the timing for you is... awkard. Just wanted you guys to be aware.

PR #207, opened 2023-04-26
feat: error checks for significant bit truncation 

ethan-crypto:
In the process of quantizing very large floating point input values of size greater than 2^k where k = 127 - b, and b is your scaling factor (2^b) used in the fixed point representation, the final value gets truncated down to 2^k.  This is significant b/c it can lead to the truncated input values being significantly smaller than they should be by up to 1/(2^(a-k) ) magnitude, where 2^a is the size of the floating point value.


alexander-camuto:
Hey @ethan-crypto ty for this 😄 

Just a note, we need to keep that inner loop parallel if possible. When loading very large weight matrices (60m+ parameters) this quantization step can take a while.

PR #211, opened 2023-04-28
feat: parallel region layouts

alexander-camuto:
Makes `Region` both `Send` + `Sync` and also lays out regions in parallel.


alexander-camuto:
closing in favour of #251

PR #217, opened 2023-05-02
feat: add percent error tolerance checks 

ethan-crypto:
Hey Dante, quick question: Other than tolerance related range checks, do you plan on using the BaseOp::Range
for anything else? If not, I can remove the range base op, since I can use the GreaterThan look ups and Sub ops to apply the same constraints as the original range_check layout.


alexander-camuto:
The benefit of leaving the Range check as is is that in the case of tol = 0 it corresponds to a (relatively) cheap equality constraint. Whereas the GreaterThan method requires an additional lookup table which is bad for circuit bloat.

src/circuit/tests.rs
```diff
@@ -1507,6 +1508,119 @@ mod rangecheck {
...
+                    },
+                )
+                .unwrap();
+            Ok(())
+        }
+    }
+
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_range_check_percent() {
```
alexander-camuto, line 1578: ty for adding these :)

src/circuit/ops/poly.rs
```diff
@@ -49,7 +49,7 @@ pub enum PolyOp<F: PrimeField + TensorType + PartialOrd> {
     Pack(u32, u32),
     GlobalSumPool,
     Iff,
-    RangeCheck(i32),
+    RangeCheck(Tolerance),
```
alexander-camuto, line 52: This is probably more suitable as a `HybridOp` as it leverage both lookups + regular constraints. See the general review for a description as to why that may be a better fit :)

src/circuit/tests.rs
```diff
@@ -1507,6 +1508,119 @@ mod rangecheck {
...
+        type Params = TestParams;
+
+        fn without_witnesses(&self) -> Self {
+            self.clone()
+        }
+
+        fn configure(cs: &mut ConstraintSystem<F>) -> Self::Config {
+            let a = VarTensor::new_advice(cs, K, LEN);
+            let b = VarTensor::new_advice(cs, K, LEN);
+            let output = VarTensor::new_advice(cs, K, LEN);
```
alexander-camuto, line 1547: Where does the `GreaterThan` lookup table get configured here ? / do these tests succeed locally from what you can tell ? You can run `cargo test circuit::tests` to run this block of tests

alexander-camuto (changes requested):
Looks super clean :) added a few comments here and there but more broadly it would be good to 
-  [ ] update the readme "general usage" section to reflect the update in the commands
-  [ ] add an integration test run in the `test_func` macro (listed below): 
https://github.com/zkonduit/ezkl/blob/699bd3992bfd388f6f76e6eb45a8886661dedbbb/tests/integration_tests.rs#L197-L222
- [ ] from what  I can tell there's currently no step which configures the lookups required for  `Tolerance::Percentage`.  Luckily we provide a method to do this on the `Op` Trait. All you need to do is make sure the `required_lookups` fn is correctly defined for the new `RangeCheck` Op. I've suggested you move it to being a `HybridOp` as it leverages both lookups and more linear constraints -- and those sorts of ops have a non default `required_lookups` fn :)

alexander-camuto:
> @alexander-camuto How did you generate the circuit file in ../tests/wasm/circuit? I need to regenerate it since I have changed the run_args field in the "ModelParams" struct. Specifically I changed the type in the tolerance field in the RunArgs struct from usize to the Tolerance enum. I believe this mismatch is what is causing the error for wasm32 tests.

Hey ! its the circuit params file from 1l_mlp: you should be able to generate it by running ezkl setup -D input.json -M network.onnx --params-path=kzg.params --vk-path=vk.key --pk-path=pk.key --circuit-params-path=circuit in that folder

alexander-camuto:
@ethan-crypto  The python binding tests are still going to break

alexander-camuto (approved):
LGTM

RusseII:
Great work!

PR #220, opened 2023-05-03
feat: python bindings for verify

JSeam2:
Changes:
1. Refactor prove for #216 to `accept circuit_params_path`
2. Create python bindings and test for verify


alexander-camuto (approved):
<3

PR #230, opened 2023-05-07
fix: issue reporting

JSeam2:
Changes:
1. Use `about` instead of `description`. This might be a reason why the issues did not show up.
2. Use lower case issue templates.
3. Allow blank issue for greater flexibility if needed.


jasonmorton (approved):
should do it

PR #254, opened 2023-05-23
perf: optimized fix_verifier_sol & added support for aggregated solidity verifiers 🎉

### LINKED ISSUE #117: aggregated proof verification in Solidity

ethan-crypto:
Reduced all of the add opcode operations in the fix_verifier_sol by precomputing the offeset memory addresses of all of the in memory variables in the solidity verifier (pubInputs, proof and transcript), adding them with their corresponding Yul bytecode addresses, and replacing the add opcodes with those results. Resolves #117

The optimizations resulted in about a 19 percent reduction in the size of the solidity verifiers. When optimizer--runs 1 is set, the solidity verifier is only 2.5 percent greater than the deployment code verifier. Whereas before the solidity verifier was about 11 percent greater. ⛽️🏌️


alexander-camuto:
Just want to check latest changes to python api match those here

src/pfsys/evm/aggregation.rs
```diff
@@ -312,13 +312,13 @@ impl Circuit<Fr> for AggregationCircuit {
...
 
-/// Create aggregation EVM verifier bytecode
+/// Create aggregation EVM verifier deployment and sol code.
 pub fn gen_aggregation_evm_verifier(
     params: &ParamsKZG<Bn256>,
     vk: &VerifyingKey<G1Affine>,
     num_instance: Vec<usize>,
     accumulator_indices: Vec<(usize, usize)>,
-) -> Result<DeploymentCode, AggregationError> {
+) -> Result<(DeploymentCode, String), AggregationError> {
```
alexander-camuto, line 321: Small thing but for clarity can we 

a) create a YulCode type which is just an alias of string `type YulCode = String` 
for legibility. 

b) we can create a gen_deployment_code function which ingests YulCode and produces DeploymentCode that way we don't have to return a tuple / it makes it clear DeploymentCode comes from YulCode. 

C) update new python.rs code accordingly.

alexander-camuto (changes requested):
Looks really really good. Just have a few cosmetic comments for code legibility + need to update the python.rs code and we should be gtg

alexander-camuto (approved):
Send it !

PR #258, opened 2023-05-26
feat: Create python bindings for `aggregate`, `verify_aggr`, `create_evm_verifier_aggr`, `print_proof_hex`

JSeam2:
Changes:
1. Creates python bindings for `aggregate`, `verify_aggr`, `create_evm_verifier_aggr`, `print_proof_hex`

Notes:
1. DeployVerifierEVM and SendProofEVM python bindings will be handled via pyezkl separately. The python bindings on the rust side should be at feature parity as of now.


alexander-camuto (approved):
LGTM !

PR #261, opened 2023-05-29
feat: build and publish python actions

JSeam2:
Changes:
1. Builds for macos x86_64, universal2
2. Builds for windows x64, x86
3. Builds for manylinux using ubuntu x86_64, i686, aarch64, armv7, s390x, ppc64le, ppc64
4. Builds for musllinux using x86_64, i686, aarch64, armv7
5. Publishes wheels to PYPI after building.

Notes:
Not entirely confident this works for all the following architectures, will need to test on the builds.

There will be a need to handle versioning properly in each PR after this, as there will be errors if conflicting versions get published to PYPI.


alexander-camuto (approved):
LFG!

PR #278, opened 2023-06-01
fix: add public/public solidity tests

ethan-crypto:
Instances in the send_proof and verify_via_solidity functions need to be flatted before they are iterated through to fill in the public_inputs vector. The previous test cases only checked the default case (public-input=false, public-output=true), which passed since the instances were only length 1 in that case. But when both public input and output are set to true the instances length is equal to 2, and therefore only the public input instances in the instances[0] were parsed into public_inputs, omiting the public_outputs which should come after.


alexander-camuto (approved):
LFG ! Will merge when tests complete

PR #282, opened 2023-06-03
fix: comment out linux-cross

JSeam2:
Changes:

1. There's a problem with the cross compilation linux toolchain for likely due to some python and gcc dependencies not being properly linked in the maturin-actions container used to compile. This PR will temporarily omit out `linux-cross` to enable builds on other platforms for the time being.


alexander-camuto (approved):
lfg !

PR #286, opened 2023-06-05
ci: update pypi.yml with testpypi upload

JSeam2:
Changes:
1. Uses a matrix instead for macos
2. Creates another step which uploads into TestPyPI if a trusted publisher setup exists there


alexander-camuto (approved):
looks good

PR #288, opened 2023-06-07
feat: wasm setup params

lancenonce:
This PR creates WASM bindings for:

Generating circuit parameters
Generating proving and verifying keys
Future work:

When WASM-bindgen multivalue is available, consolidate all three functions to one `setup()` function.


alexander-camuto (approved):
looks good !

PR #293, opened 2023-06-09
feat: Create verifyWithDataAttestation.sol

ethan-crypto:
Allows users to read inputs to their EZKL models directly on-chain and attest to them when verifying.

- Add a new parameter to the `**run args**`: **`—on-chain-inputs=true`**.
- Require **`input.json`** used by EZKL to include a field defining where to read the inputs from on-chain. 
- Define the account `address` and **`callData`** that EZKL will statically call into.
- The `decimals` field of the callData determines the number of decimal places in the on chain data to scale by. So for instance if the decimals is set to `3` for a raw on chain input value of `1111` it will get converted to `1.111` floating point value. 
- The data that is returned from these calls will be parsed into uint256 integers. Right now we only support reads into view functions that return [elementary data types](https://docs.soliditylang.org/en/v0.8.20/abi-spec.html#types).  This restriction is in place to ensure that all of returned data can be implicitly converted into uint256. If users need to read from view functions that return dynamic types they will need to create an adaptor function that parses

image


src/commands.rs
```diff
@@ -159,6 +162,10 @@ pub struct RunArgs {
     /// Flags whether params are public, private, hashed
     #[arg(long, default_value = "private")]
     pub param_visibility: Visibility,
+    /// Base used to pack the public-inputs to the circuit. (value > 1) to pack instances as a single int.
+    /// Useful when verifying on the EVM. Note that this will often break for very long inputs. Use with caution, still experimental.
```
alexander-camuto, line 166: rm this

src/commands.rs
```diff
@@ -188,6 +195,10 @@ impl Cli {
     pub fn from_json(arg_json: &str) -> Result<Self, serde_json::Error> {
         serde_json::from_str(arg_json)
     }
+    /// Create an ezkl configuration: if there is an EZKLCONF env variable, parse its value, else read it from the command line.
+    pub fn create() -> Self {
```
alexander-camuto, line 199: rm this

.github/workflows/rust.yml
```diff
@@ -144,22 +144,18 @@ jobs:
         run: cargo test --release --verbose tests::mock_public_outputs_ -- --test-threads 32
       - name: Mock proving tests (public outputs + batch size == 10)
         run: cargo test --release --verbose tests::mock_large_batch_public_outputs_ -- --test-threads 32
+      - name: Mock proving tests (public and packed outputs)
```
alexander-camuto, line 147: rm this

.github/workflows/rust.yml
```diff
@@ -144,22 +144,18 @@ jobs:
...
+        run: cargo test --release --verbose packed_tests::mock_packed_outputs_ -- --test-threads 32
```
alexander-camuto, line 148: rm this

.github/workflows/rust.yml
```diff
@@ -144,22 +144,18 @@ jobs:
...
       - name: Mock proving tests (public inputs)
         run: cargo test --release --verbose tests::mock_public_inputs_ -- --test-threads 32
       - name: Mock proving tests (public params)
         run: cargo test --release --verbose tests::mock_public_params_ -- --test-threads 32
+      - name: Mock 'everything' tests (public and packed outputs, public inputs)
```
alexander-camuto, line 153: rm this

PR #294, opened 2023-06-10
feat: move optimization runs to create evm verifier commands.

ethan-crypto:
Adds optimization runs as an optional param to both create-evm-verifier and create-evm-verifier-aggr commands.
The optimizations by the SOLC compiler are applied to the generated EVM bytecode without modifying the original solidity code. As such, it only makes sense to add another optional param, sol-byte-code-path, that determines where to output the optimized compiled solidity bytecode for deployment use later (although you can set the sol-bytecode-path without optimizations if you want).

Also added the sol_bytecode_path param to the verify-evm command in order to run tests on the sol_bytecode generated by the create evm verifier commands.

        /// The path to output the compiled Solidity bytecode
        #[arg(long)]
        sol_bytecode_path: Option<PathBuf>,
        /// The number of runs set to the SOLC optimizer.
        /// Lower values optimze for deployment size while higher values optimize for execution cost.
        /// If not set will just use the default unoptimized SOLC configuration.
        #[arg(long)]
        optimizer_runs: Option<usize>,


alexander-camuto (approved):
send it !

PR #295, opened 2023-06-11
feat: calibration pass

alexander-camuto:
Creates a separate `gen-circuit-params` command which, as the name suggests, generates the circuit params file. 

Usage is a such: 

```bash 

ezkl gen-circuit-params --calibration-data calibration-data.json --calibration-target resources --model network.onnx --circuit-params-path circuit.json 
```

Note the two new calibration related flags: 

```bash
 --calibration-data <DATA>
          The path to the .json calibration data file. If set will finetune the selected parameters to a calibration dataset

      --calibration-target <TARGET>
          Target for calibration. 
      
       Possible values:
          - resources: Optimizes for reducing cpu and memory usage
          - accuracy:  Optimizes for numerical accuracy given the fixed point representation

```

If the calibration related arguments are omitted, you can pass run-args as you would normally to simply save the file with the passed in arguments 

```bash 
ezkl gen-circuit-params --logrows=22 --scale=7 --bits=10 --model network.onnx 

```

The resulting params file can then be used by mock, forward, prove -- though note that forward and mock can still take in run_args as usual if you want to experiment rapidly. As such both 

```bash
ezkl mock --circuit-params-path circuit.json .... 
```

AND 

```bash
ezkl mock --logrows=22 --bits=21 ....
```

Though `--circuit-params-path` takes PRIORITY. 

Here's a diagram of how commands can relate to one another with this new flow: 
ezkl_flow

Major changes: 

- moving circuit params from bincode to json for human readability + ability to modify on the fly. 
- circuit params now generated BEFORE setup is called by `gen-circuit-params`
- ability to call mock and forward on circuit params output 
- reduces python.rs footprint by eliminating loading from CLI args + passing in execute.rs commands directly


src/wasm.rs
```diff
@@ -32,18 +31,16 @@ use crate::pfsys::Snarkbytes;
 /// Generate circuit params in browser
 #[wasm_bindgen]
 pub fn gen_circuit_params_wasm(
```
lancenonce, line 33: Don't we have to change the parameters here to accept a calibration folder? Or will that be included in run args?

PR #309, opened 2023-06-16
feat: hashed parameters

alexander-camuto:
- Creates a module trait
- Enables the "hashed" visibility parameter to apply to model params. 
- Removes pack base -- as its original purpose (reducing calldata) is better served by hashing
- Generalizes the input file to include parameter hashes as well.


tobinsouth:
Great feature. The (output_visibility, param_visibility, input_visibility) = (public, hashed, public) will be great for verifiability! 🥳  Do we need to change anything in calibrate-settings?

alexander-camuto:
Hey @tobinsouth :) all you need to do is set the correct visibility flags when using `gen-settings`. 
For example: 

`ezkl gen-settings --input-visibility=hashed --output-visibility=private --param-visibility=hashed`

then run  `calibrate-settings` on the generated file

PR #310, opened 2023-06-17
ci: cross releases

lancenonce:
In this PR:
- Added Cross compilation targets for `linux-gnu`, `linux-musl`, `macos-aarch64`, `macos-x86`, `win-msvc`. 
- New versions trigger assets uploaded to tag


.github/workflows/release.yml
```diff
@@ -0,0 +1,139 @@
+name: ezkl-lib-release
+on:
```
JSeam2, line 2: We should use workflow_dispatch instead

This way it helps prevent tag collisions and also adds a separate step before releasing to make sure that at least a core contributor knows what they’re releasing.

`on:
  # Requires someone to trigger this build script via github workflows
  # It is required that the codebase is audited by contributors
  # before triggering to reduce the likelihood of supply chain attacks
  workflow_dispatch: `

JSeam2: It's possible to gate builds to tagged versions with this check

```yaml
    if: "startsWith(github.ref, 'refs/tags/')"

```

.github/workflows/release.yml
```diff
@@ -0,0 +1,139 @@
...
+        
+    - name: Upload release archive
+      uses: actions/upload-release-asset@v1.0.2
+      env:
+        GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN}}
+      with:
+        upload_url: ${{ needs.create-release.outputs.upload_url}}
+        asset_path: ${{ env.ASSET }}
+        asset_name: ${{ env.ASSET }}
+        asset_content_type: application/octet-stream
```
JSeam2, line 139: Add a newline at the end of file to help git diff to differentiate between files

.github/workflows/release.yml
```diff
@@ -0,0 +1,139 @@
...
+        if: env.EZKL_VERSION == ''
+        run: |
+          echo "EZKL_VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+          echo "version is: ${{ env.EZKL_VERSION }}"
+
+      - name: Create Github Release
+        id: create-release
+        uses: softprops/action-gh-release@v1
+        with:
+          token: ${{ secrets.RELEASE_TOKEN }}
```
JSeam2, line 27: It should be possible to use `${{ github.token }}` otherwise I believe that a separate `RELEASE_TOKEN` will need to be set in the github secrets settings

.github/workflows/release.yml
```diff
@@ -0,0 +1,139 @@
...
+      shell: bash
+      run: |
+        mkdir -p build-artifacts
+        powershell.exe -Command "Compress-Archive -Path 'target/${{ matrix.target }}/release/ezkl.exe' -DestinationPath 'build-artifacts/ezkl-win.zip'"
+        echo "ASSET=build-artifacts/ezkl-win.zip" >> $GITHUB_ENV    
+        
+    - name: Upload release archive
+      uses: actions/upload-release-asset@v1.0.2
+      env:
+        GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN}}
```
JSeam2, line 134: add a space for consistency

.github/workflows/release.yml
```diff
@@ -0,0 +1,139 @@
...
+      with:
+        upload_url: ${{ needs.create-release.outputs.upload_url}}
```
JSeam2, line 136: add a space for consistency

PR #311, opened 2023-06-17
feat: visibility = encrypted 

alexander-camuto:


tests/python/binding_tests.py
```diff
@@ -126,8 +126,15 @@ def test_forward():
...
+    assert data["input_data"] == [[0.053262424, 0.074970566, 0.052355476, 0.028825462, 0.058487028,
+                                   0.008225823, 0.07530029, 0.0821458, 0.06227987, 0.024306035, 0.05793174, 0.04044203]]
+    assert data["output_data"] == [[0.053222656, 0.12841797, 0.07519531, 0.10546875, 0.20947266, 0.104003906, 0.052246094, 0.08105469, 0.028808594, 0.05859375, 0.06689453, 0.008300781, 0.13378906, 0.2241211, 0.09033203, 0.07519531, 0.15722656,
+                                    0.08203125, 0.0625, 0.08691406, 0.024414062, 0.12060547, 0.18554688, 0.064941406, 0.05810547, 0.09863281, 0.040527344]]
+    data["processed_inputs"]["poseidon_hash"] == [[8270957937025516140,
+                                                   11801026918842104328, 2203849898884507041, 140307258138425306]]
+    data["processed_params"]["poseidon_hash"] == []
```
JSeam2, line 135: There's no assert being used, so the `==` used will only lead to equality results and no effect on the pytest. 

I'll address it in #312

PR #319, opened 2023-06-21
feat: onchain / file options for input data + persisted evm state between cmds

ethan-crypto:
- Removed sol string from eth.rs. Now reads from solidity files in contracts to create the verifiers.
- The input_data and output_data fields of GraphInput now returns the DataSource enum.
- Right now this enum only contains "on-chain" and "file" options but in future PRs we can expand to include other data sources, such as databases.  
```rust
/// Enum that defines source of the inputs/outputs to the EZKL model
#[derive(Clone, Debug, Deserialize, Serialize)]
pub enum DataSource {
    /// On-chain data source. The first element is the calls to the account, and the second is the RPC url.
    OnChain(Vec<CallsToAccount>, RPCUrl),
    /// .json File data source.
    File(Vec<Vec<f32>>)
}
impl Default for DataSource {
    fn default() -> Self {
        DataSource::File(
            vec![vec![]]
        )
    }
}
```
```rust
#[derive(Clone, Debug, Deserialize, Default)]
pub struct GraphWitness {
    /// Inputs to the model / computational graph (can be empty vectors if inputs are coming from on-chain).
    /// TODO: Add retrieve from on-chain functionality
    pub input_data: DataSource, 
    /// The expected output of the model (can be empty vectors if outputs are not being constrained).
    pub output_data: DataSource,
    /// Optional hashes of the inputs (can be None if there are no commitments). Wrapped as Option for backwards compatibility
    pub processed_inputs: Option<ModuleForwardResult>,
    /// Optional hashes of the params (can be None if there are no commitments). Wrapped as Option for backwards compatibility
    pub processed_params: Option<ModuleForwardResult>,
    /// Optional hashes of the outputs (can be None if there are no commitments). Wrapped as Option for backwards compatibility
    pub processed_outputs: Option<ModuleForwardResult>
}
```

- The above change now enables modular on/off chain data sources for both input_data and output_data.
- Removed the test-reads parameter from the prove command. Replaced with optional test-on-chain-witness path for saving a new input.json that contains the OnChain data sources for input_data  and output_data.
- Also added GraphInput deseralization tests.


src/graph/input.rs
```diff
@@ -30,15 +30,30 @@ pub struct CallsToAccount {
     /// Address of the contract to read the data from.
     pub address: String,
 }
+/// Enum that defines source of the inputs/outputs to the EZKL model
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum DataSource {
+    /// On-chain data source. The first element is the calls to the account, and the second is the RPC url.
+    OnChain(Vec<CallsToAccount>, RPCUrl),
+    /// .json File data source.
+    File(Vec<Vec<f32>>)
```
alexander-camuto, line 39: you can tag this as default  using #[default] right above

src/graph/input.rs
```diff
@@ -51,82 +66,46 @@ pub struct GraphWitness {
...
-    /// Save the model input to a file
-    pub fn save(&self, path: std::path::PathBuf) -> Result<(), Box<dyn std::error::Error>> {
-        serde_json::to_writer(std::fs::File::create(path)?, &self).map_err(|e| e.into())
-    }
-}
-
-/// The input tensor data and shape, and output data for the computational graph (model) as floats.
-/// For example, the input might be the image data for a neural network, and the output class scores.
-#[derive(Clone, Debug, Deserialize, Default, Serialize)]
-pub struct GraphInput {
```
alexander-camuto, line 81: We need to bring this back

src/commands.rs
```diff
@@ -283,9 +281,9 @@ pub enum Commands {
     /// Loads model and input and runs mock prover (for testing)
     #[command(arg_required_else_help = true)]
     Mock {
-        /// The path to the .json witness file
-        #[arg(short = 'W', long)]
```
alexander-camuto, line 287: bring this back

src/commands.rs
```diff
@@ -355,9 +353,9 @@ pub enum Commands {
     /// Fuzzes the proof pipeline with random inputs, random parameters, and random keys
     #[command(arg_required_else_help = true)]
     Fuzz {
-        /// The path to the .json witness file, which should include both the network input (possibly private) and the network output (public input to the proof)
-        #[arg(short = 'W', long)]
```
alexander-camuto, line 359: and this

src/commands.rs
```diff
@@ -384,9 +382,9 @@ pub enum Commands {
     /// Loads model, data, and creates proof
     #[command(arg_required_else_help = true)]
     Prove {
-        /// The path to the .json witness file, which should include both the network input (possibly private) and the network output (public input to the proof)
-        #[arg(short = 'W', long)]
```
alexander-camuto, line 388: this

src/commands.rs
```diff
@@ -483,13 +487,13 @@ pub enum Commands {
         /// If not set will just use the default unoptimized SOLC configuration.
         #[arg(long)]
         optimizer_runs: Option<usize>,
-        /// The path to the .json witness file, which should
+        /// The path to the .json data file, which should
         /// contain the necessary calldata and accoount addresses  
         /// needed need to read from all the on-chain
         /// view functions that return the data that the network
-        /// ingests as inputs.
-        #[arg(short = 'W', long)]
```
alexander-camuto, line 491: change to -W

src/commands.rs
```diff
@@ -570,13 +574,21 @@ pub enum Commands {
...
+        /// the call data and account addresses needed to read from
+        /// evm quantized data in this file.
+        #[arg(short = 'D', long)]
+        file_witness: Option<PathBuf>,
+        /// The path to the .json data file, which should
         /// contain the necessary calldata and account addresses  
         /// needed need to read from all the on-chain
         /// view functions that return the data that the network
-        /// ingests as inputs.
-        #[arg(short = 'W', long)]
```
alexander-camuto, line 578: change to -W

src/eth.rs
```diff
@@ -152,58 +158,91 @@ fn count_decimal_places(num: f32) -> usize {
...
+            s[index+1..].len()
+        },
         None => 0,
     }
 }
 
 ///
 pub async fn setup_test_contract<M: 'static + Middleware>(
     client: Arc<M>,
-    data: &GraphWitness,
```
alexander-camuto, line 164: change to GraphWitness for clarity

src/graph/input.rs
```diff
@@ -74,19 +86,18 @@ impl GraphWitness {
...
 }
-
 /// The input tensor data and shape, and output data for the computational graph (model) as floats.
 /// For example, the input might be the image data for a neural network, and the output class scores.
 #[derive(Clone, Debug, Deserialize, Default, Serialize)]
 pub struct GraphInput {
     /// Inputs to the model / computational graph (can be empty vectors if inputs are coming from on-chain).
     /// TODO: Add retrieve from on-chain functionality
-    pub input_data: Vec<Vec<f32>>,
+    pub input_data: DataSource,
```
alexander-camuto, line 95: we def want a deserializer to that we can retain simple input.json that is either vector of floats or some on-chain calldata. OR how are you envisioning folks using this ? Is it either a vector of f32 OR a bunch of calldata ?

src/graph/mod.rs
```diff
@@ -491,11 +734,6 @@ impl Circuit<Fp> for GraphCircuit {
 
         let module_configs = ModuleConfigs::from_visibility(cs, visibility, params.module_sizes);
 
-        trace!(
-            "log2_ceil of degree: {:?}",
```
alexander-camuto, line 495: bring back

src/graph/mod.rs
```diff
@@ -511,16 +749,14 @@ impl Circuit<Fp> for GraphCircuit {
...
+        trace!("Setting input in synthesize");
         let mut inputs = self
             .inputs
             .iter()
             .map(|i| ValTensor::from(<Tensor<i128> as Into<Tensor<Value<Fp>>>>::into(i.clone())))
             .collect::<Vec<ValTensor<Fp>>>();
 
         let mut instance_offset = ModuleInstanceOffset::new();
-
-        trace!("running input module layout");
```
alexander-camuto, line 523: bring back

src/graph/mod.rs
```diff
@@ -587,7 +820,6 @@ impl Circuit<Fp> for GraphCircuit {
                 PlonkError::Synthesis
             })?;
 
-        trace!("running output module layout");
```
alexander-camuto, line 590: all of these

tests/integration_tests.rs
```diff
@@ -648,7 +674,7 @@ mod native_tests {
         let status = Command::new(format!("{}/release/ezkl", *CARGO_TARGET_DIR))
             .args([
                 "mock",
-                "-W",
```
alexander-camuto, line 651: bring back -W

tests/integration_tests.rs
```diff
@@ -714,7 +739,7 @@ mod native_tests {
                 "-M",
                 &format!("{}/{}/network.onnx", test_dir, example_name),
                 "-O",
-                &format!("{}/{}/witness.json", test_dir, example_name),
```
alexander-camuto, line 717: rename back to witness

tests/integration_tests.rs
```diff
@@ -1278,8 +1304,8 @@ mod native_tests {
         let status = Command::new(format!("{}/release/ezkl", *CARGO_TARGET_DIR))
             .args([
                 "fuzz",
-                "-W",
-                format!("{}/{}/witness_fuzz.json", test_dir, example_name).as_str(),
```
alexander-camuto, line 1282: same everywhere

alexander-camuto (changes requested):

src/graph/input.rs
ethan-crypto: Vector of f32 or a bunch of calldata at the moment.

alexander-camuto:
This PR ended up being a bit more involved than anticipated to retain backwards compatibility. 

- We now leverage a `DataSource` enum to tag inputs as being either `OnChain` or `File` based. 
    - To retain backwards compatiblity we serialize these enums as "untagged". Given some of the persistent bugs / compatibility issues between serde and serde_json with enums we have to implement our own custom serializer to do so. 
- The `DataAttestation` pipeline assumes pre-existing on-chain inputs to draw from. 
      - To simplify the cli and reduce the number of arguments being passed around we switch our EVM backend from one that spins up end-to-end sessions using anvil for the user, to one that assumes a persisted backend which is interacted with by way of an `--rpc-url` flag. 
      - A typical way to use the cli is now a) spin up anvil b) deploy contracts, verifiers etc... c) issue transaction etc... d) close anvil (persist session if you want). 
- We introduce further modularity into the DataAttestation pipeline and allow for public inputs to be provided both from on-chain and from files. 
      - you can now (for example) provide public inputs from a file, and public outputs from on-chain view sources. 
      - what we DONT enable yet is the ability to have mixed sources WITHIN public inputs (or outputs). Soon(TM). 
 - We provide a command `setup-test-evm-witness` which will populate data on-chain with inputs from a file and conversely generate a witness file with the appropriate on-chain calls to fetch that data. 
      - this is a good way to test the pipeline in fresh test envs like anvil.

alexander-camuto (approved):
gtg

PR #344, opened 2023-07-07
perf: optimized sol verifier by replacing memory params with calldata.

ethan-crypto:
- Reduced the size of the solidity verifier by setting memory location from "memory" to "calldata" and turning the dynamic size pubInputs array to a fixed sized array.
- Hardcoding the number of pubInputs in as a fixed sized array will also prevent folks from submitting pubInputs arrays of incorrect length. 
- This resulted in a 0.48 KB reduction in the size of the verifiers when solc `optimizer-runs = 1`. When the optimizations were off the reduction was even greater at 0.932 KB.


alexander-camuto (approved):
lgtm !

PR #345, opened 2023-07-07
refactor: change `ezkl_lib` to `ezkl`

JSeam2:
Changes:
1. Renames `ezkl_lib` to `ezkl`
2. Updates relevant python tests to use `ezkl` instead
3. Bumps version to `v1.2.0` to prevent collisions with the pyezkl repository

**Important Note**
Prior to building it will be necessary to swap over the trusted publisher settings in PyPI for this repo and the pyezkl repo too


tobinsouth:
What differentiates the new ezkl package (the maturin build from ezkl git) from the PyPI package ezkl? Does this now make the poetry install step of pyezkl unnecessary?

JSeam2:
@tobinsouth hey, we're going to stick with this repo as the main `ezkl` package, `pyezkl` will be called `ezkl_lib` instead. This is because we intend to decouple the main bindings from all the data science libraries being used which has been causing issues for people installing the library. 

`pyezkl` now `ezkl_lib` will now be used to provide helper functions (quantization error calculations) and visualization tools for the library.

PR #381, opened 2023-07-30
chore: add gas estimate to verify function

HadiEsna:
Add gas estimation to verify function to see how much gas units, on-chain verification uses


alexander-camuto (approved):
lgtm !

PR #384, opened 2023-07-31
fix: limit the pubinput size to verifiers to be `<` P 

ethan-crypto:
*In order to prevent the verifier from accepting two version of the same pubInput, n and the quantity (n + P),  where n + P <= 2^256, we require that all pubInputs are stricly less than P.
*The reason for this is that the assmebly code of the verifier performs all arithmetic operations modulo P and as a consequence can't distinguish between n and n + P values.

Thank you so much to @arielgabizon for flagging this issue to us
:)


alexander-camuto (approved):
LGTM !

PR #388, opened 2023-07-31
docs: check for colab env in notebooks and installs ezkl

JSeam2:
Changes:
1. Adds a check for colab environment within notebooks and installs ezkl. Otherwise, rely on local ezkl installations. This is to make the notebook examples work out of the box should someone import the notebook within colab while ensuring that the local ezkl is being used for github runner tests.


alexander-camuto (approved):
lgtm

PR #396, opened 2023-08-06
feat: add gen random elgamal variables wasm binding

ethan-crypto:


alexander-camuto (approved):
lgtm for now. Lets make a note to revisit the serialization horrors we have uncovered.

PR #418, opened 2023-08-15
feat: add solc optimizer runs param to deploy evm command

ethan-crypto:
Added back solc optimizer runs as an optional param to the deploy evm command as requested by @HadiEsna


alexander-camuto (approved):
lgtm

PR #419, opened 2023-08-16
feat: leverage sparsity throughout

alexander-camuto:


alexander-camuto:
benefit is marginal, even for large models, but adds complexity -- so am closing

PR #420, opened 2023-08-16
docs: update default demo example and update readme

JSeam2:
Changes:
1. Uses the ABCDE EZKL demo which goes through an end to end flow of training a model and deploying the model using remix
2. Updates the readme to use one-liner installation for both CLI and the Python bindings


alexander-camuto (approved):
looks good !

PR #435, opened 2023-08-25
fix: import wasm engine from mjs and cjs targets

currenthandle:
in nodejs/commonjs import from: `"main": "nodejs/ezkl.js"`
in browser/ejs import from `"module": "web/ezkl.js"`


alexander-camuto (approved):
lfg

PR #441, opened 2023-08-26
fix: bump minimum cal scale

alexander-camuto:


alexander-camuto:
closing in favour of #443

PR #481, opened 2023-09-13
feat: time series forecasting example with nbeat

JSeam2:
Changes:
1. Adds a time series forecasting example which may be useful for financial modelling related projects.


alexander-camuto (changes requested):
@JSeam2 can you clear the outputs of the new notebook ? -- will make it easier to track future changes

alexander-camuto:
just need to be added to `.github/workflows/rust.yml` as well

PR #505, opened 2023-09-28
feat: serialize/deserialize ezkljs/engine helper methods

ethan-crypto:
Added JS based serialize/deserialize helper methods to ezkljs engine package. These helper methods are meant to replicate the `serde_json::from_slice` and `serde_json::to_vec` methods used in the rust based wasm tests [here](https://github.com/zkonduit/ezkl/blob/main/tests/wasm.rs)

To convert a JS object or string representation of an engine input, (e.i. proof, witness, setting, message, e.t.c.) call the `serialized` method, then pass the results to the respective method. 

To convert an engine output to a JS object, pass it to the `deserialized` method

### DIFF
--- a/tests/wasm/testWasm.test.ts
+++ b/tests/wasm/testWasm.test.ts
@@ -41,10 +41,13 @@ describe('Generate witness, prove and verify', () => {
         const endTimeProve = Date.now();
         proof_ser = new Uint8ClampedArray(result.buffer);
         // test serialization/deserialization methods
-        const proof_ser_ref = serialize(deserialize(proof_ser));
-        const test = serialize("text");
-        console.log(test);
-        expect(proof_ser_ref).toEqual(proof_ser);
+        const proof = deserialize(proof_ser);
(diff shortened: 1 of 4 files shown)


tests/wasm/testWasm.test.ts
```diff
@@ -38,6 +40,14 @@ describe('Generate witness, prove and verify', () => {
         result = wasmFunctions.prove(witness, pk, circuit_ser, params_ser);
         const endTimeProve = Date.now();
         proof_ser = new Uint8ClampedArray(result.buffer);
+        // test serialization/deserialization methods
+        const proof = deserialize(proof_ser);
+        const proof_ser_ref = serialize(proof);
+        if (example === "prelu_gmm") {
+            expect(proof_ser_ref.length).toEqual(proof_ser.length);
```
alexander-camuto, line 47: Why does this if/else need to exist ?

alexander-camuto (approved):

PR #509, opened 2023-10-01
docs: gcn example

saeyoon17:
(WIP)
- Try enabling the training process with downsampled graphs.

Adds a GCN example.


alexander-camuto:
@saeyoon17 can you add this to the CI pipeline :) ? you need to modify `tests/py_integration_tests.py` and `.github/rust.yaml`

saeyoon17:
@alexander-camuto Sure! I’ll do that.

saeyoon17:
@alexander-camuto Hi dante, I have a question. 
I had some troubling running the tests on the local due to the dependency issues in pypi packages. 
It may be due to my mac local environment.

It does pass when I remove all specific versions but this is not desirable.. right?
Onto the original dependency, we need to add versions for torch-scatter, torch-sparse, and torch-geometric. 
For me, those were:
```
“torch-scatter==2.1.0”,
“torch-sparse==0.6.16",
“torch-geometric==2.2.0”,
```
But was not able to check the compatibility with other package versions :(.
If the versions need to be specified, would it be possible to run CI with the versions?
(If that's the case, I can revert last commit and fit in the versions.)

alexander-camuto:
Hey @saeyoon17 thanks for the question ! We should definitely keep the versions in for the CI pipeline. We've been bitten by versionless python tests in the past and I want to avoid that aha

alexander-camuto (approved):
Looks good ! tests are passing. Thank you so much. Will merge in a bit

PR #516, opened 2023-10-03
fix: manually hash set inputs for set_membership notebook

ethan-crypto:
Also added sanity check tests inside the notebook 👍


alexander-camuto (commented):
Looks good ! Can you clear output on the notebook to reduce the committed jupyter notebook code

PR #527, opened 2023-10-06
feat: custom private key

Neurone:
This PR allows the user to specify a private key to deploy the verifier contract.

This feature is meant mainly to be used with non-local Anvil networks, i.e., for test networks or main networks where the external RPC provider does not hold the users' private keys. Still, it also works with Anvil if the user wants to specify another key that is not the first one automatically generated.


alexander-camuto:
nice -- thank you for this. We had originally kept this without the private key as most folks seem to use the cli for local testing on anvil and then switch to remix or other tooling to deploy the contracts in an easier fashion. 

Is this something you had found yourself wanting to use ?

Neurone:
Hi @alexander-camuto , yep, this is something I personally use very often for testing on private or public testnets. I totally agree on the local development first, but the next step for me it's always a shared network (private or public) for integration tests before going mainnet.

alexander-camuto:
sweet let me give this a quick review.

src/python.rs
```diff
@@ -943,6 +943,7 @@ fn deploy_evm(
     sol_code_path: PathBuf,
     rpc_url: Option<String>,
     optimizer_runs: usize,
+    private_key: Option<String>,
```
alexander-camuto, line 946: can you explicitly set the default on the pyOxide binding macro above

alexander-camuto: you also need to do this for deploying the da verifier in the python bindings

Neurone: Thanks for the review, I'll take care of those bindings. Just be patient, I think I can do it during the weekend, and I'll push another commit with the updates.

alexander-camuto: awesome ty :)

Neurone: Done. I also took the opportunity to add some additional tests in `integration_tests.rs` and in `binding_tests.py` for this feature.

tests/python/binding_tests.py
```diff
@@ -499,32 +499,38 @@ def test_deploy_evm():
     assert res == True
 
 
-def test_verify_evm():
```
alexander-camuto, line 502: can we add this test back in -- it seems like we no longer test verification inside of python here

Neurone: Sure, done. Sorry, that was totally unintentional.

PR #529, opened 2023-10-06
fix: get input scales from graph for da creation

ethan-crypto:


src/execute.rs
```diff
@@ -843,18 +843,19 @@ pub(crate) fn create_evm_verifier(
...
     input: PathBuf,
 ) -> Result<(), Box<dyn Error>> {
-    use crate::graph::{DataSource, VarVisibility};
-    check_solc_requirement();
+    use crate::graph::{VarVisibility, DataSource};
 
-    let settings = GraphSettings::load(&settings_path)?;
+    check_solc_requirement();
+    let circuit = GraphCircuit::load(compiled_circuit_path)?;
+    let settings = circuit.core.settings.clone();
```
alexander-camuto, line 858: how does this change the read in scales ? 

I think what we would want to do is: 

```rust
let input_scales = circuit.core.model.get_input_scales()
```

alexander-camuto (approved):
nice -- super nice that this has simplified the ABI as well

PR #536, opened 2023-10-09
feat: evm verifier calldata wasm binding

ethan-crypto:
*implented binding + tests
*moved `halo2_solidity_verifier` dep from
`[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` to `[dependencies]`


alexander-camuto (commented):
Before proceeding -- what's the difference in wasm bundle size after the compilation of halo2_solidity_verifier ?

PR #539, opened 2023-10-11
feat: evm instances parser 

ethan-crypto:
Use this method to efficiently parse the `instances` from the halo2 evm verifier encoded calldata in your contracts that call into a verifier. For an example of how to use it, check out how its used in the data attestation verifier [here](https://github.com/zkonduit/ezkl/blob/main/contracts/AttestData.sol#L259). 

Simply pass the encoded verifier calldata to either the `getInstancesMemory` or `getInstancesCalldata` functions and they will decode the instances from the bytes. If the memory location of the bytes is calldata then call `getInstancesCalldata`, otherwise call `getInstancesMemory`. This method of parsing the instances is way more gas efficient compared to the previous implementation, as now we only need to pay memory expansion costs for copying the instances and not the entire encoded bytes array into memory (minus the 4 byte prepended function selector).


alexander-camuto (approved):
super clean. lg

PR #549, opened 2023-10-14
feat: allow for hashed and elgamal vis for DA test data

alexander-camuto:
This allows for the creation of on-chain test data that corresponds to the instances of poseidon and elgamal modules. 

Notably those instances have their quantization scales set to 1.0 and their decimal points set to 0 (as they are field elements already). 

TODO: 
@ethan-crypto currently the mulDiv for quantization fails in the case of decimal point set to 0. I wonder if there's some 0 division slipping through. Could you take a look ? 

```rust
failed: (code: 3, message: execution reverted: Math: mulDiv overflow, data: Some(String("0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000154d6174683a206d756c446976206f766572666c6f770000000000000000000000")))
```

To reproduce run `cargo nextest run kzg_evm_on_chain_input_output_hashed_prove_and_verify_::tests_on_chain_input_0_expects --no-capture`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -257,6 +257,8 @@ jobs:
         run: cargo nextest run --release --verbose tests_evm::kzg_evm_on_chain_output_prove_and_verify --test-threads 1
       - name: KZG prove and verify tests (EVM + on chain inputs & outputs)
         run: cargo nextest run --release --verbose tests_evm::kzg_evm_on_chain_input_output_prove_and_verify --test-threads 1
+      - name: KZG prove and verify tests (EVM + on chain inputs & outputs hashes)
+        run: cargo nextest run --release --verbose tests_evm::kzg_evm_on_chain_input_output_hashed_prove_and_verify --test-threads 1
       - name: KZG prove and verify tests (EVM)
         run: cargo nextest run --release --verbose tests_evm::kzg_evm_prove_and_verify --test-threads 1
       - name: KZG prove and verify tests (EVM + hashed inputs)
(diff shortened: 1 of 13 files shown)


ethan-crypto:
> This allows for the creation of on-chain test data that corresponds to the instances of poseidon and elgamal modules.
> 
> Notably those instances have their quantization scales set to 1.0 and their decimal points set to 0 (as they are field elements already).
> 
> TODO: @ethan-crypto currently the mulDiv for quantization fails in the case of decimal point set to 0. I wonder if there's some 0 division slipping through. Could you take a look ?
> 
> ```rust
> failed: (code: 3, message: execution reverted: Math: mulDiv overflow, data: Some(String("0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000154d6174683a206d756c446976206f766572666c6f770000000000000000000000")))
> ```
> 
> To reproduce run `cargo nextest run kzg_evm_on_chain_input_output_hashed_prove_and_verify_::tests_on_chain_input_0_expects --no-capture`

Fix it here: https://github.com/zkonduit/ezkl/pull/550/commits/353c091e4adc5672a94653846cff075c76ee0d85

The issue you were facing is that at deployment time, the scales where being read from the model (which was 7) And multiplying the large field element by 2^7 was enough to cause it overflow... I fixed the issue by setting the scales to 0 at deployment if visibility was hashed or encrypted. I updated the scales set in `test_from_file_data` to zero as well. We shouldn't set the scales to 1 since they are computed as 2^1 on chain.

PR #551, opened 2023-10-15
feat: adds rust hub bindings

samtvlabs:


alexander-camuto:
All the changes that were required were to implement `ToPyObject` on the structs you had created. So for instance: 

```rust
/// Stores the Artifacts
#[derive(Debug, Deserialize, Serialize)]
pub struct Artifact {
    ///stores the aritfact id
    pub id: Option<String>,
    /// stores the name of the artifact
    pub name: Option<String>,
}

impl Artifact {
    /// Export the artifact as json
    pub fn as_json(&self) -> Result<String, Box<dyn std::error::Error>> {
        let serialized = match serde_json::to_string(&self) {
            Ok(s) => s,
            Err(e) => {
                return Err(Box::new(e));
            }
        };
        Ok(serialized)
    }
}

#[cfg(feature = "python-bindings")]
impl pyo3::ToPyObject for Artifact {
    fn to_object(&self, py: pyo3::Python) -> pyo3::PyObject {
        let dict = pyo3::types::PyDict::new(py);
        dict.set_item("id", &self.id).unwrap();
        dict.set_item("name", &self.name).unwrap();
        dict.into()
    }
}
```

All we're doing here is populating a Python dict with strings etc... 

To make the conditional wasm compiles cleaner, I moved all the hub related response structs to `src/hub.rs` and then conditionally load that into lib.rs based on the target arch. 


To allow for user customization / flexibility on the deployment parameters I also adding target and run-args flags to the create-hub-artifact command: 

```rust
 #[command(name = "create-hub-artifact", arg_required_else_help = true)]
    #[cfg(not(target_arch = "wasm32"))]
    CreateHubArtifact {
        /// The path to the model file
        #[arg(short = 'M', long)]
        uncompiled_circuit: PathBuf,
        /// The path to the input json file
        #[arg(short = 'D', long)]
        data: PathBuf,
        /// the hub's url
        #[arg(short = 'O', long)]
        organization_id: String,
        ///artifact name
        #[arg(short = 'A', long)]
        artifact_name: String,
        /// the hub's url
        #[arg(short = 'U', long)]
        url: Option<String>,
        /// proving arguments
        #[clap(flatten)]
        args: RunArgs,
        /// calibration target
        #[arg(long, default_value = "resources")]
        target: CalibrationTarget,
    },

```

This allows for commands such as 

```bash
ezkl -- create-hub-artifact -A "my first model" -M examples/onnx/1l_average/network.onnx  -D examples/onnx/1l_average/input.json -O "eb477686-cd7f-4a12-aba5-c704790356b9" --input_visibility="public" --param-visibility="fixed"
```

Final step was to add a python notebook example to the CI and ... voila !

alexander-camuto (approved):

PR #569, opened 2023-10-23
fix: non default export in utils method of web wasm bundle

ethan-crypto:


alexander-camuto:
do we have confirmation this works locally ?

alexander-camuto (approved):

PR #577, opened 2023-10-26
feat: bring back gen pk and vk wasm bindings

ethan-crypto:
This PR bring back the gen pk and vk wasm bindings that were removed in this #393 

-Removed redundant test artifacts from wasm dir
-Regenerated fresh test wasm artifacts
-Added pk and vk setup to wasm
-Added tests for each.


src/wasm.rs
```diff
@@ -259,7 +268,65 @@ pub fn genWitness(
...
 
-    serde_json::to_vec(&witness).map_err(|e| JsError::new(&format!("Failed to serialize witness: {}", e)))
+    serde_json::to_vec(&witness)
+        .map_err(|e| JsError::new(&format!("Failed to serialize witness: {}", e)))
+}
+
+/// Generate proving key in browser
+#[wasm_bindgen]
+#[allow(non_snake_case)]
+pub fn genPk(
```
alexander-camuto, line 278: hey could we flip this around ? so if you look at `create_keys_wasm` it first creates the vk and then the pk as such 

```rust

    // Initialize the proving key
    let vk = keygen_vk(params, &empty_circuit)?;
    let pk = keygen_pk(params, vk, &empty_circuit)?;

```

The wasm function should be such that `genVK` takes in compiled model and srs and generates the  VK, and `genPK` then takes in the VK and params to generate the PK . 

This will be particularly useful for the stats project as it means they only need to generate the VK and not the PK + VK to get the VK (half the work)

alexander-camuto (approved):

PR #596, opened 2023-11-07
fix: Serialize cal_data into cal_data_path

Amit0617:
This feels like a mistake.


JSeam2 (dismissed):
EDIT: It looks like one of the tests are failing

alexander-camuto:
resolved more broadly by #669

PR #601, opened 2023-11-10
feat: expose print proof hex in wasm

ethan-crypto:
We should be able to print the `proof.proof` of a proof files content as a hex string instead of a uint 8 array on the client.

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -444,10 +444,10 @@ pub fn prove(
 
 /// print hex representation of a proof
 #[wasm_bindgen]
-pub fn printProofHex(proof_path: PathBuf) -> Result<String, JsError> {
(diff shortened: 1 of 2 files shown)


src/wasm.rs
```diff
@@ -442,6 +442,14 @@ pub fn prove(
         .into_bytes())
 }
 
+/// print hex representation of a proof
+#[wasm_bindgen]
+pub fn printProofHex(proof_path: PathBuf) -> Result<String, JsError> {
+    let proof = Snark::load::<KZGCommitmentScheme<Bn256>>(&proof_path)
+        .map_err(|_| PyIOError::new_err("Failed to load proof"))?;
```
JSeam2, line 449: You shouldn't be using PyIOError

alexander-camuto: probably whats preventing it from compiling

PR #604, opened 2023-11-13
chore: sequential gpu tests

samtvlabs:
This PR makes each of the GPU jobs dependent on eachother , so they dont run concurrently.


alexander-camuto (changes requested):
This makes the triggering of events conditional but they already run sequentially in that the github runner can only run 1 job at a time. 

I think most of the issues we've seen stem from the fact that `prove-and-verify-tests-gpu` runs many tests at a time. So we'd need to each sub test there with `--test-threads 1` to run in sequence

alexander-camuto:
taken care of in #632

Issue #364, opened 2023-07-17
Integration tests for EVM hardforks to validate that deployment works for different hardforks


JSeam2:
### Is your proposal related to a problem?

The evm is constantly evolving and ezkl utilizes unique opcodes for the verifier contracts. As a result, it is not certain that the produced verifier will work on every single evm hardfork. 

### Describe the solution you'd like

As a sanity check and to improve reliability, it would be useful to check what evm hardforks are supported by the generated verifier code as part of the integration tests.

As anvil is being used, it possible to explicitly specify the hardfork being used. [See anvil options](https://book.getfoundry.sh/reference/anvil/#options). The hardfork used should be explicitly declared during tests.

### Additional Information

This is related to https://github.com/zkonduit/ezkl/issues/349

glazec:
I am exploring potential solutions to this issue. One approach is to modify `tests/wasm/testBrowserEvmVerify.test.ts` so that the hardfork parameters are passed to localEVMVerify. These parameters can be obtained when the user specifies their desired fork via the CLI.

When I ran `npm run test`, I found I missed  KZG.sol to run the tests.
```
FAIL  tests/wasm/testBrowserEvmVerify.test.ts (15.519 s)
  ● localEVMVerify › should return true when verification succeeds
    ENOENT: no such file or directory, open '../ezkl/examples/onnx/1l_mlp/kzg.sol'
```

@alexander-camuto

alexander-camuto:
@glazec the best entrypoint is with cargo test as it generates the required files

@ethan-crypto can you give instructions as to how to run the test with cargo for @glazec ?

ethan-crypto:
@glazec You can run any of the cargo tests defined [here]:(https://github.com/zkonduit/ezkl/blob/main/tests/integration_tests.rs#L1000C1-L1094)

So for instance to run the `kzg_evm_prove_and_verify_` test you would run 
``` bash
cargo nextest run --release --verbose tests_evm::kzg_evm_kzg_params_prove_and_verify --test-threads 1
```
I think the best approach to this would be to spin up different anvil instances for each hardfork on different ports.

glazec:
I've successfully set up the local environment and passed the tests. Now, I'm considering how to implement tests for different forks.

Question 1: There are approximately 15 hardforks supported by Anvil:

```
Frontier,
Homestead,
Dao,
Tangerine,
SpuriousDragon,
Byzantium,
Constantinople,
Petersburg,
Istanbul,
[...]

PR #620, opened 2023-11-19
chore: add hardfork test for kzg_evm_kzg_input_prove_and_verify

glazec:
#364 

This is a sample implementation for only kzg_evm_kzg_input_prove_and_verify_ for now. Let me know what you think.

Approach:
* Add a hardfork enum
* Add a hardfork parameter for start_anvil and each test function.
* Use macro to make sure test run on each hardfork. Currently, we run London, Paris and latest blockchain fork.

Screenshot 2023-11-18 at 21 51 02

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -960,12 +960,16 @@ mod native_tests {
             ];
 
             seq!(N in 0..=16 {
-                #(#[test_case(TESTS_ON_CHAIN_INPUT[N])])*
-                fn kzg_evm_on_chain_input_prove_and_verify_(test: &str) {
+                #(#[test_case((TESTS_ON_CHAIN_INPUT[N],Hardfork::Latest))])*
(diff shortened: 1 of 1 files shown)


alexander-camuto (approved):
super clean thanks for doing this ! will merge this once the tests pass :)

ethan-crypto:
Looks good overall. I would also try to run one set of the `kzg_evm_on_chain_input_prove_and_verify` tests so that we ensure the `AttestData.sol` contract is also compatible with all the forks. So for the `kzg_evm_on_chain_input_prove_and_verify_` test we would test on the paris, london, shanghai and latest forks like we do for the `kzg_evm_kzg_input_prove_and_verify_` test.

JSeam2 (approved):
LGTM but will be good if AttestData.sol tests are included. Will accept as it is.

glazec:
Add hardfork for kzg_evm_on_chain_input_prove_and_verify_

PR #623, opened 2023-11-20
docs: add tictactoe examples

JSeam2:
Changes:
1. Adds the tictactoe examples (binary classification, anomaly detection) as part of tests.


alexander-camuto (approved):
lets gooooooooo

PR #647, opened 2023-12-05
feat: numerical accuracy reports post calibration

alexander-camuto:


tobinsouth:
Do the accuracy reports from this branch work in the meantime while waiting on fixes from tract?

alexander-camuto:
> Do the accuracy reports from this branch work in the meantime while waiting on fixes from tract?

yes for the most part

PR #659, opened 2023-12-14
Simplify the code for `impl Op<Fp> for SupportedOp`

field-worker:
There's a bunch of code duplication, I think we can simplify the implementation. As I wrote in the method doc, since each associated (inner) value of `SupportedOp` implements `Op`, an aux method to extract this inner value can be re-used across the board.
Unless, of course, you've envisioned to have the inner value change at some point / add sth that does not impl `Op`.


alexander-camuto:
yeah we originally had some things that didn't implement Op ! thanks for catching this and cleaning this up

alexander-camuto (approved):

PR #663, opened 2023-12-16
chore: update `sumpool` tensor operation docs

field-worker:
The flag was missing from arg description. Also added a doctest that shows the usage of the flag.
Just another small PR as I was going through and understanding the codebase.


field-worker:
As a sidenote, I noticed that some tensor ops are generic over `T` (e.g. `pack`) while others are tied to specific types, in this case `i128`. Do you think there'd be any value in making `sumpool` and others generic on `T`? (though I'm not mistaken, the trait bounds would have to be a little more restrictive than `T: TensorType + std::marker::Send + std::marker::Sync` to allow for multiplication, for example.

alexander-camuto:
> As a sidenote, I noticed that some tensor ops are generic over `T` (e.g. `pack`) while others are tied to specific types, in this case `i128`. Do you think there'd be any value in making `sumpool` and others generic on `T`? (though I'm not mistaken, the trait bounds would have to be a little more restrictive than `T: TensorType + std::marker::Send + std::marker::Sync` to allow for multiplication, for example.

Hey thanks for the PR. We could make it more generic -- but it would require `Div` and `Mul` if I'm not mistaken

field-worker:
That's right, also `pub fn const_div` would need to become generic on `T`. I might tackle this in a follow-up

alexander-camuto:
> That's right, also `pub fn const_div` would need to become generic on `T`. I might tackle this in a follow-up

just as a note I edited the above with 

> then it stops applying to field elements -- hence why some ops are over integers -- so as to let the end user know exactly what type works here (else a bit of a pain to debug unfortunately).

field-worker:
I'm not sure I follow - you want to be able to have a `Tensor` over field elements, or just over primitive types?
If the former, AFAIK `Field` already implements `Mul`. While `Div` trait is not directly specified in the bounds, in theory it's possible to divide two prime field elements.

But now that I look at it, I think `Div` won't be needed after all: https://github.com/field-worker/ezkl/blob/b1582bde89df7b87f52d374d8930a5dc1a6abec5/src/tensor/ops.rs#L3731

alexander-camuto:
DM me on twitter so we can discuss in full (@CamutoDante)

field-worker:
Happy to, but you'd have to msg me or follow back: https://twitter.com/field_worker99

PR #668, opened 2023-12-18
fix: prepend 0x to proof

JSeam2:
Changes:
1. prepend `0x` to the `hex_proof` for consistency and to match what the evm generally expects.


alexander-camuto (approved):
nice

PR #695, opened 2024-01-22
fix: link for CLI auto-install

tobinsouth:


tobinsouth:
Old link had PR author owner as repo owner (breaking) and didn't point to the raw script which broke the pipe to bash.

alexander-camuto (approved):
ty ty ty

PR #724, opened 2024-02-23
fix!: cleanup felt serialization language in python and wasm

alexander-camuto:


JSeam2:
This needs a `BREAKING CHANGE:` in the commit footer prior to merging to main. I have added a `!` to `fix!` as there is a breaking API Change due to change of function naming. It would be better to increment major version here as the changes here won't be compatible amongst the previous minor versions.

PR #739, opened 2024-03-12
fix: add `__version__` to python bindings

### LINKED ISSUE #692: `ezkl` python bindings do not provide `__version__` attribute
### Describe the bug
`ezkl.__version__` returns `AttributeError: module 'ezkl' has no attribute '__version__'`

### Expected behaviors
`ezkl.__version__` should return an appropriate version so users can check from within an interactive python shell what version of ezkl they are on.

jmjac:
Use version from Cargo.toml as the python module version. Resolves #692.


alexander-camuto (approved):
beautiful

PR #765, opened 2024-04-04
fix: fix @ezkljs/verify etherumjs deps

ethan-crypto:
Additionally:
-Circular dependency fix on engine web bundle.
-Separate out npm release workflow into engine and verify to deploy `@ezkljs/engine` and `@ezkljs/verify` accordingly.
-Stop tracking pnpm lock files.
-Lock in ethereumjs deps versions in `package.json`


currenthandle (approved):
lgtm

alexander-camuto:
@ethan-crypto have we checked the new flow works (bar publishing) in ur fork ?

PR #769, opened 2024-04-06
chore: update eth postgres

ethan-crypto:
Update from e2pg -> shovel


alexander-camuto (commented):
looks good -- can we check if this occasionally fails (i.e run the nb repeatedly locally or something) -- and if not add it back into the CI

alexander-camuto (changes requested):
We should use the bindings and figure out what is going on. This feels awkward

PR #774, opened 2024-04-09
Prevent multiple additions of `EZKL_DIR` to `PATH` in install_ezkl_cli.sh

winor30:
## Description
The install_ezkl_cli.sh script is supposed to add the `EZKL_DIR` to the `PATH` if it's not already present. However, due to a typo in the variable name (`EZKl_DIR` instead of `EZKL_DIR`), the script fails to correctly check if the directory is already in the PATH, and consequently, it always adds the directory to the `PATH` even if it's already there.

So this PR just fixes the typo.


alexander-camuto:
tysm for catching that

alexander-camuto (approved):

PR #778, opened 2024-04-17
feat: `kzgCommit` wasm method

ethan-crypto:


src/wasm.rs
```diff
@@ -147,6 +148,47 @@ pub fn floatToFelt(
...
+        vk.cs().degree() as u32,
+        (vk.cs().blinding_factors() + 1) as u32,
+        &params,
+    );
+
+    Ok(wasm_bindgen::Clamped(
+        serde_json::to_vec(&output).map_err(|e| JsError::new(&format!("{}", e)))?,
+    ))
+
+    // Ok(output.iter().map(|x| (*x).into()).collect::<Vec<_>>())
```
alexander-camuto, line 189: can we delete this line and then will merge

alexander-camuto (changes requested):

PR #788, opened 2024-05-05
chore: add aarch64 linux to release pipeline

### LINKED ISSUE #783: ARM support
I'm working with an Altra Q80-30 chip and reached a dead-end with a project I'm working on as it requires ezkl but is not supported for ARM (aarch64).

I gave the binary install a shot and ended up with this:

ARM architectures are not supported for Linux at the moment. If you would need support for the ARM architectures on linux please submit an issue https://github.com/zkonduit/ezkl/issues/new/choose

Thanks in advance!

alexander-camuto:
Should resolve #783 

- Swaps ethers-rs to alloy for reasons detailed here https://github.com/gakonst/ethers-rs/issues/2667


alexander-camuto:
Awaiting fixes mentioned here to progress : https://github.com/gakonst/ethers-rs/issues/2786

PR #791, opened 2024-05-06
fix: patch invalid ${{ github.ref_name#v }} syntax on verify release script

ethan-crypto:


.github/workflows/verify.yml
```diff
@@ -17,7 +17,6 @@ jobs:
   in-browser-evm-ver-publish:
     name: publish-in-browser-evm-verifier-package
     runs-on: ubuntu-latest
-    if: startsWith(github.ref, 'refs/tags/')
```
alexander-camuto, line 20: why remove the tag based trigger ?

PR #794, opened 2024-05-10
docs: add polycommit to RunArgs

JSeam2:
Changes
1. The initial docstring prompts a user to input `kzgcommit` which doesn't exist now. If a user puts in `kzgcommit` the value will default to `private` instead. Adds `polycommit` into the docstring.


alexander-camuto (approved):
beautiful

PR #818, opened 2024-06-18
feat: `create_evm_vk` python bindings 

ethan-crypto:


src/python.rs
```diff
@@ -1533,6 +1533,56 @@ fn create_evm_verifier(
...
+/// Arguments
+/// ---------
+/// vk_path: str
+///     The path to the verification key file
+///
+/// settings_path: str
+///     The path to the settings file
+///
+/// sol_code_path: str
+///     The path to the create the solidity VK
```
JSeam2, line 1547: This docstring is probably gonna confuse people. Solidity VK won’t make sense to anyone else.

src/python.rs
```diff
@@ -1533,6 +1533,56 @@ fn create_evm_verifier(
     })
 }
 
+/// Creates an Evm VK
```
JSeam2, line 1536: Would be good to eli5 about this because this will make 0 sense to an outsider

JSeam2 (approved):

alexander-camuto (approved):

PR #820, opened 2024-06-23
feat: accept valid padding in conv node

tomaz-suller:
`tract`'s representation allows no padding to be specified by setting the padding parameter explicitly to zero, or by passing `PaddingSpec::Valid`. Currently, only models defining it explicitly are supported, so this PR adds support for "valid" padding.

The original motivation behind this is adding support for [convolutional layers in Keras](https://keras.io/api/layers/convolution_layers/convolution2d/), which only allows padding to be `valid` or `same` and which, when configured to `valid`, exported to ONNX and parsed by `tract`, results in `PaddingSpec::Valid`.

### DIFF
--- a/src/graph/utilities.rs
+++ b/src/graph/utilities.rs
@@ -1161,7 +1161,8 @@ pub fn new_op_from_onnx(
             let padding = match &conv_node.pool_spec.padding {
                 PaddingSpec::Explicit(b, a) | PaddingSpec::ExplicitOnnxPool(b, a, _) => {
                     b.iter().zip(a.iter()).map(|(b, a)| (*b, *a)).collect()
-                }
+                },
+                PaddingSpec::Valid => vec![(0, 0); conv_node.pool_spec.input_channels],
                 _ => {
                     return Err(GraphError::MissingParams("padding".to_string()));
                 }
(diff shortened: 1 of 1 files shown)


tomaz-suller:
I haven't included tests since I couldn't find where to add them, but would be glad to contribute them as well with some direction

JSeam2:
@tomaz-suller the tests are found in `examples/onnx`, you can add the example onnx file and sample input there, in this case you can omit gen.py as the input will be coming from keras/tf. 

See https://github.com/zkonduit/ezkl/tree/main/examples/onnx/accuracy for an example.

Afterwards you can add the test for the example in `tests/integration_tests.rs` 

You can add the example in the following lines, you will need to increment the array lengths respectively. 
https://github.com/zkonduit/ezkl/blob/cd91f0af2668c6fbea8538d8bec153e1c12a5b9a/tests/integration_tests.rs#L203
https://github.com/zkonduit/ezkl/blob/cd91f0af2668c6fbea8538d8bec153e1c12a5b9a/tests/integration_tests.rs#L303
https://github.com/zkonduit/ezkl/blob/cd91f0af2668c6fbea8538d8bec153e1c12a5b9a/tests/integration_tests.rs#L360

Thanks for the contribution!

tomaz-suller:
The test didn't pass and I'm having a bit of a hard time debugging on VS Code since I'm on Arch and it disables installing packages outside a virtual environment using `pip`, so the test doesn't even get to run.

I'll have another look at this again later in the week, so for now I'm leaving the PR as a draft.

alexander-camuto:
@tomaz-suller unfortunately had to implement this to be able to update tract in #823 so will deprecate this PR in favor of that one.

PR #846, opened 2024-10-01
feat: iOS Bindings

ElusAegis:
### PR: Integrating iOS Bindings for the EZKL Library

This PR introduces iOS bindings for the EZKL library, developed using the [**uniffi**](https://mozilla.github.io/uniffi-rs/latest/) package. This package facilitates the generation of bindings for multiple platforms, including iOS (Swift), as well as Android (Kotlin, Java), Python, and Ruby.

### **Key Updates:**

1. Created a new `bindings` module to consolidate all bindings, moving the previously existing `python` and `wasm` bindings under it.
2. Introduced a `bindings/universal` module containing `uniffi`-compatible bindings. These bindings implement most functions by reusing code from the existing `wasm` bindings.
3. Refactored the `wasm` bindings to leverage the shared code from the `universal` bindings.
4. Added a new script `bin/ios_gen_bindings` to generate iOS bindings and package them for integration into iOS projects.
5. Included a new GitHub Actions workflow `.github/workflows/update-ios-package.yml` to test the bindings and automatically update the [iOS package](https://github.com/ElusAegis/ezkl-ios-port/blob/main/README.md) with the latest bindings when a new release of EZKL comes out.

### **Next Steps:**

We should consider migrating the [iOS EZKL package](https://github.com/ElusAegis/ezkl-ios-port/blob/main/README.md) under the **zkonduit** organization to better streamline the development process.

### **Future Considerations:**

A larger refactor could further simplify the code by avoiding the need to use `cfg(not(any(target_os = "ios", target_arch = "wasm32")))` for functions that should be excluded from the bindings. Moving such functions into a separate module would reduce the number of changes introduced by this PR and improve code clarity.


alexander-camuto:
@ElusAegis wasm tests are currently cooked because of relative imports :)

ElusAegis:
@alexander-camuto I think this is ready for another review

### Key Updates since last review:
1. Added `ios-integration-tests` to be run locally, that required to patch `uniffi_testing` crate, which has an [issue submitted](https://github.com/mozilla/uniffi-rs/issues/2269) to resolve this. This also required to add the `build.rs` file to set up correct ENV variable for the ios bindings test. Optionally, we could remove `build.rs` and instead force users to set this variable before running the tests.
2. Moved from using `target_os = "ios"` to a specific feature for ios bindings - `ios-bindings`. Additionally added `ios-bindings-test` feature for the tests. This is useful to enable some `test` dependencies only when testing, thus making the bindings leaner.
3. Restructured the `Cargo.toml` file to optimise out dependencies that are not needed for `ios` and `wasm` bindings, such as `clap`. Additionally, cleaned up the code, so that it would compile with `--no-default-features` flag, by conditinally compiling some modules only when feature `ezkl` is enabled (which was previously gated by `target_arch = "wasm32"`. This is useful when we want to have as lean version of EZKL as possible, for for example ios bindings. 
5. Fixed crate dependency issue for `wasm` tests.
5. Added `ios-integration-tests` and `swift-package-tests` to `rust.yml` to test the ios bindings and ios bindings package. 


### Next Steps:

1. To finish integration, we need to move the **[iOS EZKL Swift Package](https://github.com/ElusAegis/ezkl-ios-port.git)** under the **zkonduit** control and correctly setup repository secrets to enable the `update-ios-package.yml` to function and push releases. 

     Possibly, `update-ios-package.yml` can be integrated into `release` workflow.

alexander-camuto:
@ElusAegis a few conflicts to resolve given the reusable verifier merge and then will merge in

PR #855, opened 2024-10-27
feat: auto split onnx model (https://github.com/zkonduit/ezkl/discussions/744)

ExcellentHH:
A simple and naive solution to a discussion topic (https://github.com/zkonduit/ezkl/discussions/744) I raised.

## Script Explanation

This script addresses the challenge of generating correctness proofs for large ONNX models on machines with limited hardware capabilities. For instance, if a machine has a processing constraint of \(2^{24}\) but the model requires significantly more, it becomes difficult to handle.

To overcome this, the script automatically partitions a large model into multiple smaller sub-models based on a given upper threshold. It ensures that the intermediate results between sub-models are protected through hashing for privacy.

By splitting the large model, this approach enables verification of larger models on machines with average hardware. Additionally, it facilitates parallel validation of the models by allowing multiple sub-models to be validated simultaneously using multithreading or multiple machines, thus improving overall efficiency.

### DIFF
--- a/tools/auto_split_onnx/auto_split_onnx.py
+++ b/tools/auto_split_onnx/auto_split_onnx.py
@@ -1,0 +1,296 @@
+import onnx
+import networkx as nx
+import numpy as np
+import json
+from onnxsim import simplify
+import os
+import ezkl
+import math
+from collections import deque
+import matplotlib.pyplot as plt
+import time
+from collections import Counter
+import argparse
+
+def find_closest_power(total_value):
+    n = math.ceil(math.log2(total_value))
+    return n
+
+def generate_random_input_data(onnx_model_path, data_path):
+    # Load the ONNX model
+    model = onnx.load(onnx_model_path)
+    graph = model.graph
+    
+    # Get the input information of the model
+    input_info = []
+    for input_tensor in graph.input:
+        # Get the dimensions of the input; assume all dimensions are known
+        dims = [dim.dim_value for dim in input_tensor.type.tensor_type.shape.dim]
+        input_info.append(dims)
+        
+    # Generate random data and store it as JSON
+    data = {'input_data': []}
+    for dims in input_info:
+        # Generate random data matching the input dimensions
+        random_input = np.random.rand(*dims).reshape([-1]).tolist()
+        data['input_data'].append(random_input)
+    
+    # Serialize the data to a file
+    with open(data_path, 'w') as json_file:
+        json.dump(data, json_file)
+
+
+def load_onnx_model_to_graph(onnx_model_path):
+    model = onnx.load(onnx_model_path)
+    graph = model.graph
+    ops = [node.name for node in graph.node]
+    # print(ops)
+    G = nx.DiGraph()
+    for node in graph.node:
+        # print(node.name)
+        inputs = [node.input[0]] if node.op_type != 'Concat' and node.op_type != "Add" and len(node.input) > 0 else node.input
+        G.add_node(node.name, op_type=node.op_type, inputs=inputs, outputs=[o for o in node.output])
+        for input_name in inputs:
+            G.add_edge(input_name, node.name)
+        for output_name in node.output:
+            G.add_edge(node.name, output_name)
+            
+    return G, ops
+
+
+def process_graph(simplified_model_path, G, ops, upper_bound_per_subgraph):
+    # Record the input and output lines of the operations
+    ops_io = {op: {"inputs": [], "outputs": []} for op in ops}
+    for op in ops:
+        ops_io[op]["inputs"] = list(G.predecessors(op))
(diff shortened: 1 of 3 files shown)


alexander-camuto:
@ExcellentHH thanks so much for getting this over the finish line. Will review this in a bit but first off huge congratulations on seeing this through 🎉🎉🎉🎉🍾🍾🍾🍾🍾

jasonmorton:
Great stuff indeed! I think we're going to want to change to KZG commitments rather than hashes to save rows, and add a final loop to actually compute the proofs, glue, and verify in an integration test.

JSeam2:
Amazing work! One usability caveat that might potentially cause issues are networks with recurrent structures. 

The following line should deal fine with DAG type of networks
```
    # Topological sorting
    topo_sorted_nodes = list(nx.topological_sort(G))
```

For better usability it might be worth flagging cycles in the network to users, and provide an error message saying the scheme will not support such kinds of networks.

tools/auto_split_onnx/auto_split_onnx.py
```diff
@@ -0,0 +1,296 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Process ONNX model and generate subgraphs.")
+    parser.add_argument("--onnx_model_path", type=str, default='./resnet18.onnx', help="Path to the ONNX model. Default is './resnet18.onnx'.")
+    parser.add_argument("--simplified_model_path", type=str, default='./resnet18_simplified.onnx', help="Path to save the simplified ONNX model. Default is './resnet18_simplified.onnx'.")
+    parser.add_argument("--upper_bound_per_subgraph", type=int, default=23, help="Upper bound per subgraph in the form of 2^n. Default is 23.")
+    parser.add_argument("--input_shape", type=str, default='{"input": [1, 3, 224, 224]}', 
+                        help="Input shape for the ONNX model in JSON format. Default is '{\"input\": [1, 3, 224, 224]}'.")
+    parser.add_argument("--simplify", action='store_true', 
+                        help="Flag to indicate if the model should be simplified. Default is False.")
+
```
alexander-camuto, line 279: can we add an argument to specify the visibility of the stitched commitments eg. `public`, `hashed`, `polycommit`

tools/auto_split_onnx/auto_split_onnx.py
```diff
@@ -0,0 +1,296 @@
...
+    parser = argparse.ArgumentParser(description="Process ONNX model and generate subgraphs.")
+    parser.add_argument("--onnx_model_path", type=str, default='./resnet18.onnx', help="Path to the ONNX model. Default is './resnet18.onnx'.")
+    parser.add_argument("--simplified_model_path", type=str, default='./resnet18_simplified.onnx', help="Path to save the simplified ONNX model. Default is './resnet18_simplified.onnx'.")
+    parser.add_argument("--upper_bound_per_subgraph", type=int, default=23, help="Upper bound per subgraph in the form of 2^n. Default is 23.")
+    parser.add_argument("--input_shape", type=str, default='{"input": [1, 3, 224, 224]}', 
+                        help="Input shape for the ONNX model in JSON format. Default is '{\"input\": [1, 3, 224, 224]}'.")
+    parser.add_argument("--simplify", action='store_true', 
+                        help="Flag to indicate if the model should be simplified. Default is False.")
+    
+    args = parser.parse_args()
```
alexander-camuto, line 280: can we add a flag to (optionally) compile the circuit (or not) -- but if the `upper_bound_per_subgraph` is 23 logrows for eg. it should 1. calibrate 2. reduce the logrows to 23 manually 3. compile

tools/auto_split_onnx/auto_split_onnx.py
```diff
@@ -0,0 +1,296 @@
...
+    onnx.save(model_simp, output_path)
+    
+    return check
+
+
+def check_total_assignments(json_file, n):
+    is_pass = False
+    with open(json_file, 'r') as f:
+        data = json.load(f)
+        total_assignments = data.get("total_assignments", 0)
```
alexander-camuto, line 176: this should be checking `num_rows` not `total_assignements`

alexander-camuto (changes requested):
This is a great start for this. I've added some comments. 

It would also be great to add an integration test for all this that splits a smaller model like nanoGPT into 2^17 chunks, generates the witness for each one, proves all the chunks, then verifies that the proof commitments match using the method from the example notebooks :)

ExcellentHH:
@alexander-camuto @jasonmorton @JSeam2 Thank you for the response and encouragement. Special thanks to @alexander-camuto for the improvement suggestions—I’ve learned a lot from them, and I apologize for the immaturity of my code. I’ve been a bit busy recently, but I’ll work on the code improvements as soon as possible. Thanks again!

PR #886, opened 2024-12-10
fix: ios bindings update action 

ElusAegis:
1. Fix an issue to allow to push new tags to the `ezkl-swift-repo` even if there are no changes to the bindings
2. Fix an issue with workflow failing on automatic trigger by removing the automatic trigger

### DIFF
--- a/.github/workflows/update-ios-package.yml
+++ b/.github/workflows/update-ios-package.yml
@@ -89,6 +89,8 @@ jobs:
           git config user.name "GitHub Action"
           git config user.email "action@github.com"
           git remote set-url origin https://zkonduit:${EZKL_SWIFT_PACKAGE_REPO_TOKEN}@${{ env.EZKL_SWIFT_PACKAGE_REPO }}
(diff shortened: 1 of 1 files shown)


ElusAegis:
Right now we need to manually trigger the workflow to publish new version of swift package. 

What we can do is add an automatic trigger for the workflow instead of the manual trigger that will run the workflow every time a new tag is created in the ezkl repo. Is this something we would prefer instead of manual updates?

alexander-camuto:
> Right now we need to manually trigger the workflow to publish new version of swift package.
> 
> What we can do is add an automatic trigger for the workflow instead of the manual trigger that will run the workflow every time a new tag is created in the ezkl repo. Is this something we would prefer instead of manual updates?

yes plz @ElusAegis

ElusAegis:
@alexander-camuto now the workflow runs when the new version tag is pushed

alexander-camuto (approved):

PR #973, opened 2025-04-24
feat: vka hashing

ethan-crypto:


.github/workflows/static-analysis.yml
jasonmorton, line 1: Let's get Zizmor to pass before merging

PR #997, opened 2025-07-31
fix: correct commitment type in ipa_commit function

GarmashAlex:
Fix logical error in ipa_commit function - use Commitments::IPA instead of Commitments::KZG when getting SRS path to ensure correct SRS file type is used for IPA commitments.


alexander-camuto (approved):
tyvm

PR #1003, opened 2025-08-12
fix: map Utf8Error and FromHexError to JsError in wasm hex decoder

Galoretka:
Replace ? and JsError::from in utf8_bytes_to_hex_decoded with explicit map_err conversions to JsError, ensuring proper error propagation under wasm32.


alexander-camuto (approved):
lgtm man

PR #1021, opened 2026-03-04
Add ONNX example: PLA Sigmoid (replace lookup with piecewise linear)

changshenhan:
This adds an ONNX example under `examples/onnx/pla_sigmoid` that replaces Sigmoid lookup with a piecewise linear approximation (PLA). The circuit uses only mul/add/compare (copy and arithmetic constraints), no lookup.

  - Keeps accuracy ≥ 99.5% vs true sigmoid.
  - Allows lighter ezkl settings (`num_inner_cols=1`, minimal `lookup_range`).
  - Useful as a reference for reducing proof cost when handling nonlinear ops.

  Author: @changshenhan


changshenhan:
Hi @jasonmorton and @zkonduit-team,

I've submitted this PR to implement the Piecewise Linear Approximation (PLA) for Sigmoid, which significantly optimizes the proof generation time.

It's been a few days and the CI workflows are still awaiting approval. Could you please help to run the workflows so I can ensure everything passes on your side?

I've already verified the logic locally with a precision of 99.5%. Thanks for your time!

PR #1028, opened 2026-05-17
fix(python): initialize and expose ICICLE GPU runtime

peter941221:
# Summary

This fixes the Linux Python GPU path for `ezkl`.

Before this patch:

- the CLI path initialized the ICICLE backend/device through `run()`
- Python `gen_srs`, `setup`, and `prove` bypassed that path
- with `ICICLE_BACKEND_INSTALL_DIR` set, Python could still stay on CPU
- once Python was forced onto the CUDA path, backend loading could still fail because the ICICLE frontend shared libraries were not globally visible to later `dlopen` calls

After this patch:

- `gen_srs`, `setup`, and `prove` all initialize the GPU backend/device through the shared execution path
- Python module initialization promotes the required ICICLE frontend shared libraries to `RTLD_GLOBAL` on Linux GPU builds
- CUDA backend loading can resolve ICICLE symbols correctly

# Root Cause

There were two linked issues.

1. Control-flow gap

- `set_device()` existed, but only the CLI `run()` path called it
- Python reached GPU-sensitive halo2 / ICICLE code directly

2. Dynamic-linking gap

- after fixing the control-flow issue, CUDA backend loading still depended on symbols exported by:
  - `libicicle_device.so`
  - `libicicle_hash.so`
  - `libicicle_field_bn254.so`
  - `libicicle_curve_bn254.so`
- in Python, those frontend libraries were not globally visible by default
- later backend `dlopen` calls could fail to resolve symbols like `register_deviceAPI`

# Changes

## `src/execute.rs`

- make `set_device()` idempotent with `Once`
- call `set_device()` from:
  - `gen_srs_cmd`
  - `setup`
  - `prove`

## `src/bindings/python.rs`

- on Linux GPU builds, promote the ICICLE frontend shared libraries to `RTLD_GLOBAL` during module initialization

# Local Validation

Validated locally on `2026-05-17` in WSL on an RTX 5090.

## Build

```text
cargo check --lib --no-default-features --features python-bindings,gpu-accelerated,ezkl
maturin develop --no-default-features --features python-bindings,gpu-accelerated,ezkl
```

## Python repro

Minimal Python chain:

1. `gen_settings`
2. `compile_circuit`
3. `gen_witness`
4. `setup`
5. `prove`

Observed behavior:

- before patch:
  - Python logged only `Registering DEVICE: device=CPU`
- after only the execution-path fix:
  - Python entered CUDA setup, but backend loading still depended on global ICICLE symbols
- after the full patch:
  - Python logs `Registering DEVICE: device=CUDA`
  - backend loads successfully
  - `setup` and `prove` both succeed

# Scope

This is intentionally narrow:

- no algorithmic proving changes
- no non-Linux Python linker changes
- no unrelated build-system or dependency changes

### DIFF
--- a/src/bindings/python.rs
+++ b/src/bindings/python.rs
@@ -28,10 +28,70 @@ use pyo3_stub_gen::{
 };
 use snark_verifier::util::arithmetic::PrimeField;
 use std::collections::HashSet;
+#[cfg(all(target_os = "linux", feature = "gpu-accelerated"))]
+use std::ffi::{CStr, CString};
 use std::str::FromStr;
+#[cfg(all(target_os = "linux", feature = "gpu-accelerated"))]
+use std::os::raw::{c_char, c_int};
 use std::{fs::File, path::PathBuf};
 
 type PyFelt = String;
+
+#[cfg(all(target_os = "linux", feature = "gpu-accelerated"))]
+#[allow(unsafe_code)]
+fn promote_icicle_frontend_libs_global() -> Result<(), String> {
+    const ICICLE_FRONTEND_LIBS: [&str; 4] = [
+        "libicicle_device.so",
+        "libicicle_hash.so",
+        "libicicle_field_bn254.so",
+        "libicicle_curve_bn254.so",
+    ];
+    const RTLD_NOW: c_int = 2;
+    const RTLD_NOLOAD: c_int = 4;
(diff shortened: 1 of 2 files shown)


peter941221:
Small follow-up: this is still the same focused Python ICICLE runtime initialization fix. If you want it split further or narrowed before review, I can do that as a minimal follow-up.

PR #1030, opened 2026-05-25
Fix Tensor::move_axis for higher-rank transposes

npow:
Off-by-one in the coordinate-remap loop. When tract decomposes a 6D Transpose like `perm=[0,2,4,1,3,5]` into a sequence of `AxisOp::Move` ops, the `source > destination` branch with `i > source` writes `old_coord[i-1]` instead of `old_coord[i]`, so the inverse index lands out of bounds and `pyo3` surfaces it as a `PanicException` during `gen_settings`.

Affects any model with a non-trivial higher-rank Transpose (transformer-style architectures all hit it).

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -946,6 +946,12 @@ impl<T: Clone + TensorType> Tensor<T> {
     /// let mut expected = Tensor::<IntegerRep>::new(Some(&[1, 3, 5, 2, 4, 6, 7, 9, 11, 8, 10, 12]), &[2, 2, 3]).unwrap();
     /// let b = a.move_axis(2, 1).unwrap();
     /// assert_eq!(b, expected);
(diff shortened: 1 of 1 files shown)


npow:
Closing in favor of two focused replacement PRs:

- #1031: Fix Tensor::move_axis for higher-rank transposes (the bug fix, minimal change to src/tensor/mod.rs only)
- #1032: Add --ignore-lookup-range-check and verbose calibration diagnostics (the feature, rebased cleanly against current main)

Splitting because maintainers prefer small focused PRs, and a crash bug is much easier to review and land in isolation from a feature addition. The diagnostics feature also needed a clean reimplementation against the substantially-refactored upstream (src/python.rs moved to src/bindings/python.rs, calibrate() signature changed, only_range_check_rebase removed, etc.).

